Issue #39, opened 2021-10-28
Add cookie banner


rosswyatt:
New component has been added for cookie banner

https://design-system.service.gov.uk/components/cookie-banner/

rosswyatt:
Closing as now have something in

PR #47, opened 2021-11-02
#39

rosswyatt:
Add cookies banner to the functions


rosswyatt:
passes all builds and checks and seems to work as intended, so solves issue #47

Issue #40, opened 2021-10-28
Add Accordion


rosswyatt:
Could add in the accordion component:

https://design-system.service.gov.uk/components/accordion/

rosswyatt:
so reading this, will need to design some js to make this work as intended

rosswyatt:
Added through PR #48

PR #48, opened 2021-11-02
#40

rosswyatt:


rosswyatt:
Added something that seems to solve #48 and the accordion.  No issues on the checks and builds

Issue #43, opened 2021-10-28
Add Tab


rosswyatt:
Could add in the tab component:

https://design-system.service.gov.uk/components/tabs/

rosswyatt:
Think this is working with PR #50

PR #50, opened 2021-11-11
#43

rosswyatt:
Add tabs as a function.  Solves at least the starter of issue #43


rosswyatt:
Passes tests and checks so good to go i think

PR #53, opened 2021-12-22
Error summary

df-just:
Added the error summary function to the package. I haven't added functionality for making the text links due to the way shiny dynamically assigns address paths on run. Can into how to introduce this later if it'd be helpful.


rosswyatt:
added a way of updating the error summary and added to the example.

rosswyatt:
This seems to pass the check and looks to be working now.  Maybe needs to be revisited to see if can be automated more, but I think good to go for now.

PR #56, opened 2022-02-07
Update gds v4.0.0

rosswyatt:
Updated to v4.0.0 of GDS' front end and fixing breaking changes in https://github.com/alphagov/govuk-frontend/releases/tag/v4.0.0


rosswyatt:
This looks to be working for me.  It passes tests, checks and builds so shouldn't cause any issues.  Most changes are just css changes anyway.

PR #57, opened 2022-02-08
adding new layout functions

rosswyatt:


rosswyatt:
This seems to pass all checks etc, so good to merge I think

PR #79, opened 2024-05-24
Fix backlink styling and update beta banner example

cjrace:
### Overview

After updating the CSS to v5.4 the backlink wasn't styling right and the tag in the beta banner now looked a bit odd as it's now longer defaulting to all caps.

### Backlink

I changed this to an action link instead, which seemed to help it pick up the styling. I tested on the example app and functionality did seem to work fine, though the tests were failing as the placing of the label within the underlying object had moved - so I've updated them to match the updates to the function.

I also added a `label` argument so that people can customise the label if they want, though left the default as "Back", so in theory it stays backwards compatible.

Before:
image

After:
image

### Beta banner

This was a quick one to update the case in the examples to be in line with updated GDS tag component

Before:
image

After:
image

### Documentation

I had a newer version of roxygen2 locally so updated the package to that to then update the docs, assuming this hasn't caused any issues?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -12,7 +12,7 @@ Description: Collection of 'shiny' application styling that are the based on
 Depends: R (>= 3.1.0)
 License: GPL-3
 Encoding: UTF-8
-RoxygenNote: 7.1.2
+RoxygenNote: 7.3.1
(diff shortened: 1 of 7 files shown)


sarahmwong (approved):
Changes look good to me thanks!

PR #82, opened 2024-06-05
Fix small size of header

sarahmwong:
Previously looked like this where the header text was too small Screenshot 2024-06-05 at 15 23 34

Was pointing to old css element that has now been renamed.

Updated here:
Screenshot 2024-06-05 at 15 26 30


jenisayshi (approved):
also happy size is responsive to screen and media query works well

Issue #60, opened 2024-04-18
Update to GDS Frontend v5+


cjrace:
The [GDS frontend is currently at v5.3.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.3.0), this package is currently on v4.something.

v5 was a big step change for a chunk of things. We should update the components in this package to match - expecting it will be a fairly big job to go through and check them all individually, so worth chunking up once the initial update of the CSS etc is on a branch.

sarahmwong:
Bits I've spotted that we should update - please add if I've missed anything!

- [ ] App title header text looks to small compared to department name
- [ ] Radio buttons - centralise text when inline = TRUE
- [ ] File upload - looks like default shiny styling, needs to point to right css element?
- [ ] Back link - needs styling - may just need to point to correct css element
- [ ] Inactive tag - need to add to run_example to test
- [ ] Warning text - currently prints out "Warning" after the inputted text in the function - think this is an error
- [ ] Cookie banner not working - this is a separate issue #67

cjrace:
Add a few extra bits I think we should check to the list above. Might be worth adding extra tabs to the example dashboard to test them with more easily?

For the cookie banner not working I think we should potentially just leave that for the separate issue to pick up?

And then final mop up steps

- [ ] Document process for updating CSS
- [ ] Update css_changes.md file
- [ ] Update version number
- [ ] New release to CRAN?

sarahmwong:
Thanks Cam! I'll message the assignees in our teams and randomly assign bits to check/see if they're happy. Agreed cookies one can be removed and treated as separate issue. Would we want to merge to dev first along with the other changes in the issue log so we only need to do one CRAN release? Or if you think these changes are more urgently needed happy to see what we need to do to release it sooner 😄

cjrace:
Let's merge to dev first and then see where we're at?

I think I might have just resolved two of the ones I added already as I got a bit carried away and started testing things out straight away! The beta banner and back link ones I'll make a quick PR pointing at to the update-to-gds-v5.4 branch shortly for you / someone else to take a look at 😄

rmbielby:
Here's the before and afters for govTable and govTabs. As far as I can see govTable looks fine, govTabs has the border line going all the way round instead of leaving a gap where the tab label is. If anyone else wants to play spot the difference and sees anything else that's not formatted as before, let me know!

[...]

PR #86, opened 2024-06-14
GDS radio button update

rmbielby:
Covering off the radio button label centre alignment for inline buttons in accordance with Issue #60.

image

Note that it centre aligns within the overall container and not to the width of the options, so requires careful use by anyone using inline as it is. The remit says to centre align the text, but maybe it makes more sense to centre align the options as well as the text when inline?


sarahmwong:
Thank you Rich! Sorry my comment wasn't very clear at all on the original issue - centre aligning the options (as in the text next to the buttons!) was what I'd meant - happy if the actual label text is left-aligned if this isn't too much of a pain to revert!

This fix looks good though, and think fixes the issue we have with text being underneath the radio buttons in the current version - however I can't replicate your output in the run_example - does it need any other arguments added to it? Still looks like this for me:

Screenshot 2024-06-17 at 11 50 41

rmbielby:
I'm ditching this one due to misunderstanding what was being asked for. Seemed easier after that to start afresh on a separate branch.

PR #87, opened 2024-06-17
Gov summary

jenisayshi:
Added **gov_summary** to the **run_example** function at the bottom of the Feedback page - 

image

**Issue** - words split when wrapping across smaller screen sizes

image


sarahmwong (approved):
Thanks Jeni - happy that this fixes the initial ask of adding the summary tab to run example. Please could you add the issue you've found here to the issue log? Note that it is an issue in the current CRAN version of ShinyGovStyle too

PR #88, opened 2024-06-17
Check tag inputs text

jenisayshi:
Updated text case and included correct text case and all colours in run_example

image

### DIFF
--- a/R/run_example.R
+++ b/R/run_example.R
@@ -126,17 +126,17 @@ run_example <- function(){
             heading_text("tag_Input", size = "s"),
 
             shinyGovstyle::tag_Input("tag1", "Default"),
-            shinyGovstyle::tag_Input("tag1", "Grey", "grey"),
-            shinyGovstyle::tag_Input("tag1", "Green", "green"),
-            shinyGovstyle::tag_Input("tag1", "Turquoise", "turquoise"),
-            shinyGovstyle::tag_Input("tag1", "Blue", "blue"),
-            shinyGovstyle::tag_Input("tag1", "Light-blue", "light-blue"),
-            shinyGovstyle::tag_Input("tag1", "Purple", "purple"),
-            shinyGovstyle::tag_Input("tag1", "Pink", "pink"),
-            shinyGovstyle::tag_Input("tag1", "Red", "red"),
-            shinyGovstyle::tag_Input("tag1", "Orange", "orange"),
-            shinyGovstyle::tag_Input("tag1", "Yellow", "yellow"),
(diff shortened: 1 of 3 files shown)


tests/testthat/test-tag.R
```diff
@@ -1,6 +1,6 @@
 test_that("tag works", {
 
-  tag_check <- shinyGovstyle::tag_Input("tag1", "COMPLETE")
+  tag_check <- shinyGovstyle::tag_Input("tag1", "Complete")
   expect_identical(
     tag_check$children[[1]],
     "COMPLETE"
```
sarahmwong, line 6: check here fails - probably just needs changing to lower case here but run devtools::test() to make sure tests pass

R/run_example.R
```diff
@@ -124,10 +124,22 @@ run_example <- function(){
...
+            shinyGovstyle::tag_Input("tag3", "Green", "green"),
+            shinyGovstyle::tag_Input("tag4", "Turquoise", "turquoise"),
+            shinyGovstyle::tag_Input("tag5", "Blue", "blue"),
+            shinyGovstyle::tag_Input("tag6", "Light-blue", "light-blue"),
+            shinyGovstyle::tag_Input("tag7", "Purple", "purple"),
+            shinyGovstyle::tag_Input("tag8", "Pink", "pink"),
+            shinyGovstyle::tag_Input("tag9", "Red", "red"),
+            shinyGovstyle::tag_Input("tag10", "Orange", "orange"),
+            shinyGovstyle::tag_Input("tag11", "Yellow", "yellow"),
+            shinyGovstyle::tag_Input("tag12", "Incorrect colour", "violet"),
```
sarahmwong, line 139: Is this testing that violet defaults to Default instead? If so think we can remove from this bit and maybe add this into a unit test instead as we'll want to catch any changes to this in future

sarahmwong (changes requested):
Thanks Jeni! The merge is blocked as tests are failing - you can run devtools::check() before pushing to make sure all is working as expected (cam's set it up so that github will automatically run this before it lets us merge anything into important branches which is v clever!)

R/run_example.R
jenisayshi: I have removed this from run example but wasn't sure how to include in test that? I had a look at the object in the environment and I'm not sure you can actually pull out the colour (navy)

sarahmwong: ooo yes good spot you're right it'd need to be a UI shinytest I think, which we've not got in there at the moment. I'll raise as a more general issue but happy if you want to merge 😄

sarahmwong (approved):

PR #90, opened 2024-06-27
Updated alignment of radio button item labels

rmbielby:
This should be more in line with what was asked for. I've set the alignment properly now so that radio button item labels are centre aligned vertically with their parent radio button.

The screenshot below shows what it looks like now after some small css tweaks:

image

Note that I've also added an extra check into the run_example() code as I don't think the stacked radio items were aligning properly previously either and I wanted to see both in action.


sarahmwong (approved):
Thank you Rich this looks great!

PR #94, opened 2024-07-19
govTabs gds54 update

rmbielby:
## Main edits

govTabs weren't rendering properly, so this is a quick update to the css to get them back on track:

- The class previously was .js-enabled, but this seems to have switched to .govuk-frontend-supported now. I've updated all instances of `js-enabled` as the class in the R code for `govTabs to govuk-frontend-supported`
- To force a gap in the border between the tab label and the tab for the selected tab, the previous css used a -1px bottom margin to shift the label border to override the tab border. To get the same effect now needs a margin of -2px. Bit random (i.e. I don't know why it's -2px rather than -1px now), but seems to work.

## Other edits

I've added `govTable` and `govTabs` examples to the example dashboard run using `run_example()`

### DIFF
--- a/R/run_example.R
+++ b/R/run_example.R
@@ -138,7 +138,7 @@ run_example <- function(){
           value = "panel3",
           gov_layout(
             size = "two-thirds",
-            backlink_Input("back1"),
+            backlink_Input("back2"),
             heading_text("Page 3", size = "l"),
             label_hint("label3", "These are some examples of using tabs and
                        tables"),
@@ -147,7 +147,9 @@ run_example <- function(){
               "tab1", example_data, "Test", "l", num_col = c(2,3),
               width_overwrite = c("one-half", "one-quarter", "one-quarter")),
             heading_text("govTabs", size = "s"),
-            shinyGovstyle::govTabs("tabsID", data, "tabs")
+            shinyGovstyle::govTabs("tabsID", data, "tabs"),
+            heading_text("button_Input", size = "s"),
+            button_Input("btn4", "Go to next page"),
           )
         ),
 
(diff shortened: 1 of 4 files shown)


R/run_example.R
```diff
@@ -121,13 +134,32 @@ run_example <- function(){
 
         #####################Create third panel################################
         shiny::tabPanel(
-          "Feedback Types",
+          "Tables and tabs",
           value = "panel3",
           gov_layout(
             size = "two-thirds",
-            backlink_Input("back2"),
+            backlink_Input("back1"),
```
cjrace, line 141: Should this be back2? Currently it jumps all the way back to the first tab

R/run_example.R
```diff
@@ -121,13 +134,32 @@ run_example <- function(){
...
             heading_text("Page 3", size = "l"),
-            label_hint("label3", "These are some examples of the types of user
+            label_hint("label3", "These are some examples of using tabs and
+                       tables"),
+            heading_text("govTable", size = "s"),
+            shinyGovstyle::govTable(
```
cjrace, line 146: The table title is coming out in grey - don't know if it's for this PR or another one, but should be a quick CSS change to make it black? 

GDS component
image

`run_example()` on this branch
image

R/run_example.R
```diff
@@ -121,13 +134,32 @@ run_example <- function(){
...
+          )
+        ),
+
+        #####################Create feedback panel################################
+        shiny::tabPanel(
+          "Feedback Types",
+          value = "panel-feedback",
+          gov_layout(
+            size = "two-thirds",
+            backlink_Input("back2"),
```
cjrace, line 160: Should this be back3? Seems odd it jumps to tab2 from tab4

cjrace (changes requested):
Few minor comments @rmbielby

R/run_example.R
rmbielby: Fixed in the css

R/run_example.R
rmbielby: Should also be updated. And added a button on panel3 to go to the fourth tab as well.

cjrace (approved):

PR #95, opened 2024-08-16
Accordions update

jenisayshi:
update accordians js to match gds 5.4


sarahmwong:
Thanks Jeni for adding comments and making js changes! All looks good to me

PR #96, opened 2024-08-16
Update to gds v5.4

sarahmwong:
Updates all elements to GDS v5.4 - full details found in #60 

Specific fixes include:
- Fixing the show/hide all elements of accordions
- Fixing the formatting of the View Cookies link
- Adding elements to run_example() for easier checking


cjrace:
Think we should probably leave #60 open until this is on the main branch @sarahmwong? Though all looks great aside from that

cjrace (approved):

Issue #64, opened 2024-04-18
Add an example (or function) for adding a skip to main content link


cjrace:
Skip to main content links are incredibly valuable for users with assistive software, we should add an example of one into this package, either in the template app or as a new function.

sarahmwong:
Closed by #101

PR #97, opened 2024-08-22
Created skip_to_main function

TRENDS-Rachel-Avbulimen:
I have created the function "skip_to_main" to generate the accessible "Skip to main content" link. Please let me know if I need to amend or change anything in the code/script.

This solves issue #64

### DIFF
--- a/R/run_example.R
+++ b/R/run_example.R
@@ -28,6 +28,7 @@ run_example <- function(){
       cookieBanner("Run Example"),
       header("MOJ", "ShinyGovstyle Example",
              logo="shinyGovstyle/images/moj_logo-1.png", logo_width = 66),
+      shinyGovstyle::skip_to_main(),
       banner(
         "banner",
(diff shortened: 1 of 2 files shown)


sarahmwong:
Thanks for this Rachel! 

I'm running this on my mac but can't get the skip link to appear in either the run_example or the example provided in the function. Not sure if this is just a mac issue, but you can test out by doing devtools::load() and then running run_example() to check :) 

I think some of the CMD checks are also failing as the documentation just needs updating - if you run devtools::document() then that should update the .Rd files too. Possibly calling in the package name in the run_example script is causing it to fall over too as this function isn't in the latest CRAN version of the package? I'm not sure though, but would remove just to be sure!  You can run devtools::check() which should replicate the checks that are run on github locally too to make sure they pass :D

sarahmwong:
closed - second PR #101 addresses this issue

PR #100, opened 2024-09-04
Sidebar nav

jenisayshi:


R/contents_links.R
```diff
@@ -0,0 +1,146 @@
...
+#'       shiny::tags$div(
+#'         shiny::tags$h2("Contents"),
+#'         contents_links(
+#'           "Feedback types",
+#'           "feedback_types_button",
+#'           subcontents_text_list =  c("tag_Input", "details")
+#'         ),
+#'
+#'         contents_links(
+#'           "Tables, tabs and accordians",
```
cjrace, line 100: I think it should be 'accordions' instead of 'accordians'? Not sure if this is misspelt elsewhere too, might be worth searching across the repo for it?

R/contents_links.R
```diff
@@ -0,0 +1,146 @@
...
+#'           subcontents_text_list = c("govTable", "govTabs", "button_Input"),
+#'           subcontents_id_list = c(NA, NA, "button_input_tables_tabs_accordians")
+#'         ),
+#'         contents_links("Cookies", "cookies_button")
+#'       )
+#'     )
+#'   )
+#'
+#'   server <- function(input, output, session) {
+#'
```
cjrace, line 111: For the example is it worth showing the way you should use observeEvent in the server to change the panels based on the links being clicked?

cjrace (changes requested):
Couple of quick comments plus:

1. Could we add an example to the readme with a screenshot for how to do this with the sub page links (and without)?
2. A bit cheeky, but while adding an example to the readme would you have time to review and update some of the other screenshots too? I know the accordions have changed in appearance now, and the tags have too, there might be more
3. When you click on one of the main contents links (not the sub ones) the page focus jumps up slightly jarringly to the very top of the page, is there a way to make it focus your screen to the main header of the page you've clicked to go to instead? (if that makes sense?)

cjrace (approved):

PR #101, opened 2024-09-12
add skip_to_main function

jenisayshi:
Added skip to main function 
- not sure if example in function is working? copied over from run_example but button doesn't seem to work here? 
- image not loading again in read me! not sure what's going on here - Sarah would you be able to take a screenshot and add to read me? :)


sarahmwong:
Thanks @jenisayshi this is looking really good! Please can you add your manual CSS changes to the bottom of the log file css_changes.md so we know what to update for future rounds?

PR #104, opened 2024-10-28
value box using colours from govuk tags

AndyABaker:


sarahmwong:
Thanks @AndyABaker! 

- [ ] - The checks are currently failing here, I think it's because a pound sign is included in the run_example value box and the container that runs the test doesn't handle that well (although this would work locally/when the package is made available!). Please could you make a change to use a different prefix and run devtools::check()/push up again? 
- [ ] - Please could you add your manual css changes and a description of what they are to the css_changes.md file? That will ensure we know what to update for the next GDS style update :)
- [ ] - Please could you run devtools::check() and push up the resulting .Rd files? Those will update the function and package documentation to fit the changes you have made.
- [ ] - If you could also add an example to the README, that would be brilliant!

@cjrace just to note the behaviour of the £ sign with the run example tests atm! Probably not worth trying to fix as it's just for the testing environment?

sarahmwong:
> @AndyABaker @sarahmwong Just been using the `value_box` function and have a quick comment / query. I can't see what the inputId is needed for, so just wondering whether it is needed and if so what it's function is?

Thank you for spotting Rich - I don't think we need this variable so happy to remove! Will raise a new issue if you're happy to action? This hasn't gone out in CRAN release yet so shouldn't be a breaking change for users if we get rid of this one

PR #107, opened 2024-11-14
fix accordion misnumbering

jenisayshi:
Fix accordion function to work for numbers greater than 9 and adjust run_example to include accordion with > 9 sections


sarahmwong:
Thanks Jeni looks good, happy with the fix and thank you for spending time thinking about how to test this properly - removed the extra accordions from the example for neatness but happy that functionality is fixed and can be maintained from your unit tests

PR #111, opened 2024-12-05
fix bug where govtable falls over if width_overwrite = NULL

### LINKED ISSUE #78: govTable function fails if width_overwrite=NULL is used
image

jenisayshi:
Wrapped the custom table width code to only run if width_overwrite is not null and added tests. Fixes #78


sarahmwong (changes requested):
Thanks Jeni! Fix looks good to me, thank you for adding tests too. Please can we add a line in at the top of the function documentation to clarify that if using width_overwrite, the specified widths for columns need to add up to one? Will just need the R documentation running off again and then happy to merge

cjrace (changes requested):
Think you've got one small conflict in README, that will need fixing before merging, and I've left one super nitpicky suggestion on spacing in the documentation. If you accept it and commit in GitHub you'll still need to pull and then rerun the documentation to update it. 

Otherwise it all seems to work well, am I right in thinking this PR fixes https://github.com/dfe-analytical-services/shinyGovstyle/issues/78? Gave it a quick test and it seems to 😄 if you reference that in the PR description like 'Fixes #78' then it should automatically link and then close that issue when this PR is merged in

sarahmwong (approved):

cjrace (approved):

PR #114, opened 2024-12-06
Allow footer links

### LINKED ISSUE #77: Add ability to add links into the footer
There's a few examples of this starting to appear, e.g. on [Explore education statistics](https://explore-education-statistics.service.gov.uk/) we add a line about OSR and also links to various pages from the footer.

image

The [UKHSA dashboard](https://ukhsa-dashboard.data.gov.uk/) also does the same.

image

We can use the govuk-footer__inline-list class for this, and should make sure to give the option for a visually hidden heading that screen readers can use to navigate to these.

When we do this change, we should also add `.noWS = "after"` to the a tag for Open Government Licence as you can now remove the excess whitespace it adds pretty easily.

Example of current footer with whitespace is visible on the README
image

cjrace:
## Overview of changes

Resolves #77, by adding an extra argument to the footer, allowing a vector of link text to be added, automatically generating actionLinks and IDs that can be used to control a hidden tabset.

Gone for this approach as I think it balances simplicity for the users with the functional need. As R Shiny is almost always a single page app, I assume most people would be using actionLinks to trigger a tabset shifting rather than actual href's to external pages. If we think this is a need, then I'll need to think again about the argument users can use. An alternative could be to allow users to put in any shiny tagList they deem appropriate, would be easier for us but then opens the door to more weird and wonderful stuff that might not be ideal or have the right styling.

Have added a functional example to the `run_example()` app too.

image

## PR Checklist

## Reviewer instructions

1. Does it work as you'd expect?
2. Is the documentation intuitive to follow?
3. Is there enough test coverage?
4. Should we be allowing standard anchor links in this as well? (if so, then probably best to do that now before merging)
5. Do we still need the `full = TRUE / FALSE` option? Having poke at it a bit, I can see difference in the output code, but can't work out the difference for an end user of a dashboard? If we don't, I could try to tidy that up in this PR too?
6. Should the crown symbol be on the right? Just noticed in my screenshot it wraps underneath, which feels odd, that wasn't something I've consciously done!

### DIFF
--- a/R/footer.R
+++ b/R/footer.R
@@ -65,7 +65,7 @@ footer <- function(full = FALSE, links = NULL) {
           shiny::div(
             class = "govuk-footer__meta-item govuk-footer__meta-item--grow",
             if (!is.null(links)) {
-              div(
+              shiny::div(
                 # Set a visually hidden title for accessibility
                 shiny::h2(
                   class = "govuk-visually-hidden",
@@ -84,6 +84,21 @@ footer <- function(full = FALSE, links = NULL) {
           shiny::tagList(
             shiny::div(
               class = "govuk-footer__meta-item govuk-footer__meta-item--grow",
+              if (!is.null(links)) {
+                shiny::div(
+                  # Set a visually hidden title for accessibility
+                  shiny::h2(
+                    class = "govuk-visually-hidden",
+                    "Support links"
+                  ),
+                  shiny::tags$ul(
+                    class = "govuk-footer__inline-list",
+
+                    # Generate as many links as needed
+                    lapply(links, footer_link)
+                  )
+                )
+              },
               shiny::tag("svg", list(
                 role = "presentation",
                 focusable = "false",
--- a/R/footer.R
+++ b/R/footer.R
@@ -34,7 +34,7 @@
 #'   shinyApp(ui = ui, server = server)
 #' }
 #'
-#' footer(links_list = c("Footnotes", "Support", "Accessibility statement", "Cookies"))
+#' footer(links = c("Accessibility statement", "Cookies"))
 footer <- function(full = FALSE, links = NULL) {
   # Validation on the links input
   if (!is.null(links)) {
--- a/R/footer.R
+++ b/R/footer.R
@@ -1,21 +1,29 @@
 #' Footer Function
 #'
 #' This function create a gov style footer for your page
+#'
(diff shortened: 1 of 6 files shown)


sarahmwong (commented):
Thanks Cam!

1. Works as expected
2. documentation straightforward and easy to follow (Example in the function doesnt have an accessibility page to click into, dont htink thats a problem though as shows the layout/the cookie link functions fine
3. tests are sufficient
4. happy to only include internal shiny app links like you suggest - dont see why there would be a need for external links but could be one to monitor if anyone raises
5. Happy to get rid of full argument
6. Yes think image should be on the right if possible! Happy to pick up if this is a faff and you are heading off for xmas soon!

cjrace:
> Thanks Cam!
> 
> 1. Works as expected
> 2. documentation straightforward and easy to follow (Example in the function doesnt have an accessibility page to click into, dont htink thats a problem though as shows the layout/the cookie link functions fine
> 3. tests are sufficient
> 4. happy to only include internal shiny app links like you suggest - dont see why there would be a need for external links but could be one to monitor if anyone raises
> 5. Happy to get rid of full argument
> 6. Yes think image should be on the right if possible! Happy to pick up if this is a faff and you are heading off for xmas soon!

Have changed my mind on 5 as when I checked back I found the difference (and not sure why I didn't first time around!) - have added an image on the README to show it more clearly.

Had a quick look and couldn't immediately see where I've messed up the image going onto a new line - happy for you to look if you want @sarahmwong, and if not I'm happy to have a look when I'm back around the 23rd. Aware there's some fun conflicts too, happy to pick them up when I'm back as it's with my other PR

sarahmwong:
Thanks Cam! Compared to master and think a div class was missing causing it to fall out of line with the rest of the footer! Should be sorted in last commit now 😄

sarahmwong (approved):

cjrace:
> Thanks Cam! Compared to master and think a div class was missing causing it to fall out of line with the rest of the footer! Should be sorted in last commit now 😄

🤦 it's always something simple isn't it! Thanks for figuring that out and fixing!

cjrace:
Got into a knot locally with VSCode and merge conflicts so the history on this branch is a bit of a mess, hoping once squashed and into master it's all okay, though this is definitely a place to double check if we have any issues over the next few weeks @sarahmwong!

PR #118, opened 2024-12-11
add shinytest2 for testing run example and draft yaml for deploy

### LINKED ISSUE #123: Accordions only open when you click "hide" or "show"
## Overview
Accordions only open when a user directly clicks "hide" or "show".

Expected behaviour is that the accordions open and close when a user clicks anywhere on the accordion

## How to reproduce
Steps to reproduce the behaviour:

1.  Run devtools::load_all or library(shinyGovstyle)
2. Run run_example()
3. Nav to "tables, tabs and accordions"
4. Click on the accordion to try and open it
5. Accordion won't open until you click "show"

## Expected behaviour
Expect accordion to open when you click anywhere on the accordion

## Additional context
FIrst raised by Cam in #117

cjrace:
## Overview of changes

Prepared `run_example()` so we can deploy it on our shinyapps.io account.

Updated the run_example test to use shinytest2, which opens up the potential for more UI testing of components.

Fixed a bug in `run_example()` (found by the shinytest2 tests spitting out a warning) where we had two IDs with the same name:

image

While I was there I used styler on the `R/run_example.R` file, I think that accounts for most of the changes - just code formatting rather than substantive.

## PR Checklist

## Reviewer instructions

Mostly just check the changes in `run_example()` app. If that's fine then I think this is good to in.

Once I've got the deployment working I'll update the docs to point to it, waiting until it's working to do that though.

I've prepared a draft yaml file for the deployment GitHub action, expecting I may need to do some trial and error on the master branch once this is merged to get it working, I've never deployed an app returned by a function in this way before (alternative could be to make an app folder within the inst/ directory and then make run_example() call shiny::runApp("inst/showcase_app") (or something to that effect. I'll try to make this way work first, and then if I need to make bigger changes to move code around I'll do it in a new PR.

### DIFF
--- a/R/run_example.R
+++ b/R/run_example.R
@@ -1,6 +1,6 @@
 #' Example Function
 #'
-#' This function runs a shiny example using different parts of the package
+#' This function runs a shiny example showcasing different parts of the package
 #' @return a shiny app with examples in
 #' @keywords example
 #' @export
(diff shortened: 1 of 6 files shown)


R/run_example.R
```diff
@@ -1,457 +1,458 @@
...
-             logo="shinyGovstyle/images/moj_logo-1.png", logo_width = 66),
+      header("MoJ", "shinyGovstyle showcase",
+        logo = "shinyGovstyle/images/moj_logo-1.png", logo_width = 66
+      ),
       banner(
         "banner",
         "Beta",
-        'This is a new service \u002D your <a class="govuk-link" href="#">
-        feedback</a> will help us to improve it.'),
+        'This is a new service \u002D your <a class="govuk-link" href="https://github.com/dfe-analytical-services/shinyGovstyle/issues/new/choose">
```
sarahmwong, line 39: could use external link function here once we have it merged in

R/run_example.R
```diff
@@ -1,457 +1,458 @@
...
+                    "Recently updated",
+                    "Most views",
+                    "Most comments"
+                  ),
+                  select_value = c("published", "updated", "view", "comments")
+                ),
+                heading_text("file_Input", size = "s"),
+                file_Input(inputId = "file1", label = "Upload a file"),
+                heading_text("button_Input", size = "s"),
+                button_Input("btn1", "Go to next page")
```
sarahmwong, line 176: takes you to halfway down feedback page, expect to be taken to text types instead

R/run_example.R
```diff
@@ -1,457 +1,458 @@
...
+                text_area_Input(
+                  inputId = "text_area2",
+                  label = "How are you today?",
+                  hint_label = "Leave blank to trigger error",
+                  error = T,
+                  error_message = "Please do not leave blank",
+                  word_limit = 300
+                ),
+                heading_text("button_Input", size = "s", id = "button_input_text_types"),
+                button_Input("btn2", "Go to next page"),
```
sarahmwong, line 215: go to next page doesn't do anything here (also replicates the button we have on the previous page - could remove?)

R/run_example.R
```diff
@@ -1,457 +1,458 @@
...
   Cases_open <- c(3, 1, 2, 24, 16, 24, 98, 122, 126, 1380, 1129, 1539)
   Cases_closed <- c(0, 0, 0, 18, 20, 27, 95, 131, 142, 1472, 1083, 1265)
   data <- data.frame(tabs, Case_manager, Cases_open, Cases_closed)
 
   shiny::shinyApp(
     ui = shiny::fluidPage(
-      title="ShinyGovstyle",
-      cookieBanner("Run Example"),
+      title = "ShinyGovstyle",
+      cookieBanner("shinyGovstyle showcase"),
```
sarahmwong, line 31: can't get "view cookies" to work here

sarahmwong (changes requested):
Thanks Cam! Teeny bugs spotted here - think these have just slipped through when we've been adding things over time - shiny tests will be a really neat way to avoid this in future 😄

R/run_example.R
cjrace: Yeah, just removed as I don't think we need, we have the other example of it

R/run_example.R
cjrace: We had a few conflicting observes so made it go to text types now (though haven't sorted it going to the top of the page, it just stays how far you'd scrolled down on the page)

R/run_example.R
cjrace: Definitely!

R/run_example.R
sarahmwong: I'll raise as an issue!

Issue #110, opened 2024-12-04
Deploy a live version of the example dashboard


cjrace:
## What?
Currently in the package there is the `run_example()` function, as well as an extensive list of screenshots in the README, we could host this somewhere so users of the package can view it at a URL rather than needing to run it themselves.

## Why?
I think having a link to a hosted version of the example app would be a nice to have for potential users to browse the components in action and make more use from the code we have for `run_example()` - I suspect more people will browse a live app than run the example themselves. 

After doing this we could also then move the screenshots out of the README and into a separate vignette on the documentation site with a floating navigation to make it easier to navigate or move some of the information into the documentation on the individual components themselves. 

I think this would also help when we get towards developing a template, as it'll act as a showcase app for available components, allowing the scope of the template app to be smaller and more focussed.

## How?
I'd be happy for DfE to host this for ease, as we already have pipelines that can deploy a dashboard to our shinyapps.io account (to be moved to POSIT Connect servers in early 2025).

cjrace:
Still failing, though seems like we are getting closer! This is the action log to look at for fixing - https://github.com/dfe-analytical-services/shinyGovstyle/actions/workflows/example-app-deploy.yaml

PR #120, opened 2024-12-11
Move run_example to own directory

cjrace:
## Overview of changes

After #118 failed, I'm now going for the bigger change of creating a separate directory for the `run_example()` app, so I have a directory I can point shinyapps.io to for the deployment. Hoping this approach should work...

Relates to #110 but won't necessarily resolve it - will need to check that after it goes into master.

`run_example()` still runs the example app.

## PR Checklist

## Reviewer instructions

I copied the contents of run_example into an app.R file in it's own directory so nothing material should have change in the example app (aside from the note around where to find the code).

Main thing is to check that run_example() still opens an app for you, and that the package checks pass


sarahmwong (changes requested):
Thanks Cam! App runs fine with run_example() and devtools::check() passes all tests for me. 

devtools::test() I get some errors though - cant find the skip_to_main() function, also skips the tests for testing the example app, can see it's meant to skip on CRAN but is this expected behaviour here? I have also just updated R on my mac so may be a me problem! I'll let you know if that turns out to be the case in this thread, but do you get the same errors? image

cjrace:
Rebased onto the main branch and resolved conflicts, force pushing the updates - will look into your comment now @sarahmwong - I don't think that is expected!

cjrace:
I think I've worked it out @sarahmwong 🤞  `devtools::check()` builds the package again from scratch and tests that package, whereas `devtools::test()` doesn't, it just tests it. I was getting the same as you, and then I ran 
```
remotes::install_github("moj-analytical-services/shinyGovstyle")
```
to reinstall the package on my laptop generally, after that, the tests passed - I think the changes in the footer function had been made since I'd last installed so that was causing the error. In theory, if you reinstall the package from GitHub and then run again it should work fine with either `devtools::test()` or `devtools::check()`

sarahmwong:
Thanks Cam! You're right both now work :D Weirdly says it skips the example_app test but then goes ahead and runs example_app basic_load test so think skip_on_cran is working as expected - happy to approve image

sarahmwong (approved):

Issue #127, opened 2025-01-06
GDS compliant Download function


rmbielby:
## Describe the solution you'd like
I haven't been able to find anything specific to download interactivity in the GDS style guide. Looking around government services, downloads tend to be links, e.g.

<img width="466" alt="image" src="https://github.com/user-attachments/assets/cc9a7e94-7399-4be4-b1ad-c9859096eb8e" />

So key elements to the solution would probably be:
- Link has relevant text (with validation to prevent inappropriate text, e.g. "here") 
- Link has appropriate html / labelling
- Link has GDS type link styling
- Following text with options to show file type and size where known
- The link triggers a script, i.e. `shiny::downloadHandler()`, that can run some R code to create a file to download and then trigger the download

The styling basis and file size text could probably just be lifted from `dfeshiny::external_link()`.

## Describe alternatives you've considered
DfE dashboard creators have so far been using the default shiny download button. This got the following write-up from an accessibility review:

> The ‘Download data’ link has conflicting attributes applied to it which might be confusing for assistive technology.
Currently the `<i>` element found within the link has the ‘role="presentation"’ attribute applied which should programmatically hide the element from assistive technology, however due to the implementation of the ‘aria-label’ attribute this overrides the attribute which means it is discoverable to assistive technology. As a result of this the link will have the accessible name of ‘download icon Download data’.
Also, there is an ‘aria-live’ attribute applied which should only be used to inform assistive technologies about regions of a web page that are dynamically updated and should be announced to users in a non-interruptive manner. In its current form this does not provide screen reader users with a status message, however the inclusion of this attribute is not suitable for this scenario.
Solution:
Remove the ‘aria-label’ and the ‘aria-live’ attribute as they are not required. The ‘role="presentation"’ attribute will inform assistive technologies that an element should be treated as purely presentational and should not be announced to the user. 
Removing the ‘aria-label’ will provide the link with an accessible name of ‘Download data’ which is expected.

There doesn't seem to be much of a government precedence for using a button, just a link seems more appropriate.

cjrace:
Secondary function we could add for this, and should do before closing the issue is a download radio options module - discussed in #128 

There is another type of download pattern that we could account for - the radio button file type selection with a confirmation button to download, like how EES does it in the table tool? ([example featured table, you'll see if you scroll to the bottom](https://explore-education-statistics.service.gov.uk/data-tables/fast-track/4bf00b2f-dec4-43da-ec12-08dcdd324466))
[...]

PR #128, opened 2025-01-06
Download link function

rmbielby:
## Overview of changes

Adding a standard way to show downloadable content in the form of `download_link()`. I've based it heavily on a combination of `shiny::downloadLink()` and `shinyGovstyle::external_link()`. As with `shiny::downloadLink()`, it works with `shiny::downloadHandler()`, but had the added options to explicitly provide the file type and file size.

This is the resulting html when using `download_link()`:

`<a id="download_data" class="shiny-download-link shiny-bound-output" href="session/1492a1e83012ff241b242c018e6b28b6/download/download_data?w=" target="_blank" download="" aria-live="polite">Download data (CSV, 1 KB)</a>`

Which looks like this in the app (note the **download_link** is just the heading in the demo app):

image

Note that `aria-live="polite"` is still set, but I think this is the right way to go. For example if the user set's the link_text or file_size to update automatically based on some dynamic selections, then we'd want that to be announced at the "the next graceful opportunity".

I've not included a dynamic way to do the file sizes in this, but user's can just do their own workaround by using a reactive that evaluates a predicted file size (e.g. `table_size_in_mb <- reactive(...)`) and then calling the download link as `download_link("table_data", "Download table data", file_size = table_size_in_mb())`. Admittedly the `...` in the above is doing a lot of heavy lifting in that description, but to be fair, shinyGovstyle is about styling rather than writing active code for user's, so maybe fulfilling that part would be overreach anyway.

Fixes some of #127 

## PR Checklist


R/download_link.R
```diff
@@ -0,0 +1,130 @@
+#' Download link
+#'
+#' @param outputId The name of the output slot that the `downloadHandler` is assigned
+#' to.
+#' @param link_text Text that will appear describing the download action
+#' (default: "Download file").
```
cjrace, line 6: Default looks to be "Download data" not "Download file"?

R/download_link.R
```diff
@@ -0,0 +1,130 @@
...
+#'       paste0(input$dataset, ".csv")
+#'     },
+#'     content = function(file) {
+#'       # Write the dataset to the `file` that will be downloaded
+#'       write.csv(data(), file)
+#'     }
+#'   )
+#' }
+#'
+#' shinyApp(ui, server)
```
cjrace, line 44: To be consistent with elsewhere I'd wrap in `if(interactive()) {}` instead of `\dontrun{}` (e.g. header.R)

R/download_link.R
```diff
@@ -0,0 +1,130 @@
...
+#' If the string is shorter than 7 characters a console warning will be thrown.
+#' There is no way to hush this other than providing more detail.
+#' @param file_type The file type to be download (default: CSV)
+#' @param file_size The file size if known. Needs to be a string ending in one
+#' of KB, MB, GB or rows.
+#'
+#' @returns shiny.tag object
+#' @export
+#'
+#' @examples
```
cjrace, line 18: I think it would be worth including an example showing the file size (either in the example you have, or as an additional one)

R/download_link.R
```diff
@@ -0,0 +1,130 @@
+#' Download link
+#'
```
cjrace, line 2: Could we add a note explaining the importance of including the file size?

R/download_link.R
```diff
@@ -0,0 +1,130 @@
...
+#'       write.csv(data(), file)
+#'     }
+#'   )
+#' }
+#'
+#' shinyApp(ui, server)
+#' }
+download_link <- function(
+    outputId,
+    link_text = "Download data",
```
cjrace, line 48: Does this not fail our own bad links checks? (if not, should it? It doesn't feel particularly descriptive)

tests/testthat/test-download_link.R
```diff
@@ -0,0 +1,61 @@
...
+test_that("Rejects dodgy link text", {
+  expect_error(download_link("download_data", "Click here"))
+  expect_error(download_link("download_data", "here"))
+  expect_error(download_link("download_data", "PDF"))
+  expect_error(download_link("download_data", "Full stop."))
+  expect_error(download_link("download_data", "http://shiny.posit.co/"))
+  expect_error(download_link("download_data", "www.google.com"))
+})
+
+test_that("Surrounding whitespace shrubbery is trimmed", {
```
cjrace, line 33: image

R/download_link.R
```diff
@@ -0,0 +1,130 @@
...
+    ))
+  }
+
+  # Assuming all else has passed, make the link text a nice accessible link
+  # Note the file_size is manually assigned and optional right now. Ideally,
+  # this would be updated dynamically when linking to a dynamically created
+  # file, such as the CSV version of a table in an app.
+  if (!is.null(file_size)) {
+    if(!grepl("KB$|MB$|GB$| rows$", file_size)){
+      error("File size should be a string ending in one of KB, MB, GB or rows.")
```
cjrace, line 102: Is the error function from a package? Could we explicitly state it 
image

Or did you mean `stop()`? Actually, I think you did probably mean `stop()` as that's how we do this elsewhere?

inst/example_app/app.R
cjrace, line 1: Worth adding a section for downloads into the README? Currently we have examples of most of our functions in there as well as the example app (though we could also decide to cut back from the README and lean on the app more if it feels too duplicative)

cjrace (changes requested):
Nice! Honestly quite a neat solution 😄 

To be fair, I've just looked around all over to try to find a reference point for this, and I can't find one anywhere either. So given that, I think I agree with your comment about shinyGovstyle mainly being for styling, so with a few tweaks to the function documentation I'd say this PR could fully resolve #127.

One other thought I'd had when looking through is that there is another type of download pattern that we could account for - the radio button file type selection with a confirmation button to download, like how EES does it in the table tool? ([example featured table, you'll see if you scroll to the bottom](https://explore-education-statistics.service.gov.uk/data-tables/fast-track/4bf00b2f-dec4-43da-ec12-08dcdd324466))
image

I have done a version of this in the [provider dashboard in DfE](https://github.com/dfe-analytical-services/apprenticeships-provider-dashboard/blob/main/R/dashboard_modules/05-national_provider_summary.R), which mainly uses shinyGovstyle components, so interested to know if you think that's worth wrapping up in a bigger module in a separate PR, or if there's enough parts already we could just add an example of it in this PR to the showcase app with existing functions? Either way I think it'd be a pattern worth adding to this package to show how it should look (and allow us to easily test it in our upcoming accessibility audit)

rmbielby:
> Nice! Honestly quite a neat solution 😄
> 
> To be fair, I've just looked around all over to try to find a reference point for this, and I can't find one anywhere either. So given that, I think I agree with your comment about shinyGovstyle mainly being for styling, so with a few tweaks to the function documentation I'd say this PR could fully resolve #127.
> 
> One other thought I'd had when looking through is that there is another type of download pattern that we could account for - the radio button file type selection with a confirmation button to download, like how EES does it in the table tool? ([example featured table, you'll see if you scroll to the bottom](https://explore-education-statistics.service.gov.uk/data-tables/fast-track/4bf00b2f-dec4-43da-ec12-08dcdd324466)) image
> 
> I have done a version of this in the [provider dashboard in DfE](https://github.com/dfe-analytical-services/apprenticeships-provider-dashboard/blob/main/R/dashboard_modules/05-national_provider_summary.R), which mainly uses shinyGovstyle components, so interested to know if you think that's worth wrapping up in a bigger module in a separate PR, or if there's enough parts already we could just add an example of it in this PR to the showcase app with existing functions? Either way I think it'd be a pattern worth adding to this package to show how it should look (and allow us to easily test it in our upcoming accessibility audit)

I like the idea of the radio-download module, but instinct is that it should be a separate PR. Happy to do that work (especially as you've already done most of the leg-work in the provider dashboard anyway!), but think it's big / distinct enough it warrants it's own review process.

R/download_link.R
rmbielby: Done

R/download_link.R
rmbielby: It's a fair cop. I've added `download data` to the banned list now and removed any default as well. If we're wanting to avoid generic text, I guess it's better not to give users an easy way out with a default value.

inst/example_app/app.R
rmbielby: Added a small section. I would definitely vote in favour of deleting pretty much all of the example function reference from the readme though and just have people use the auto-generated reference. Feels like we're double maintaining.

R/download_link.R
rmbielby: I've removed the default now anyway, based on a later comments

R/download_link.R
cjrace: So good you did it twice...

inst/example_app/app.R
cjrace: Raised #129 to cover this separately

cjrace (changes requested):
All looks great, raised #129 and added a comment on #127 for the extra things to look at in other PRs. 

One final thing, as the file size thing still doesn't quite sit right, I think I've worked out what's missing - can we add a warning if the file_size argument is blank? That way if it is blank there's a very clear prompt to add it, and if anyone so desires they can ignore too (though really if they're using this it's probably because they care about best practices so they should expect to be nudged on them)

rmbielby:
> All looks great, raised #129 and added a comment on #127 for the extra things to look at in other PRs.
> 
> One final thing, as the file size thing still doesn't quite sit right, I think I've worked out what's missing - can we add a warning if the file_size argument is blank? That way if it is blank there's a very clear prompt to add it, and if anyone so desires they can ignore too (though really if they're using this it's probably because they care about best practices so they should expect to be nudged on them)

I've added a console warning now (and some tests).

cjrace (approved):
Thanks for adding that!

PR #131, opened 2025-01-10
Softened rules around alt text in header from errors to warnings

rmbielby:
## Overview of changes

The alt text that was added in the header was becoming a bit of an awkward breaking change, whilst we're managing versions between dfeshiny and shinyGovstyle. I've softened the rules from errors to warnings for now to give a bit of a grace period, whilst dashboards are getting up to speed.

## PR Checklist


cjrace (approved):
All works as expected, happy this makes it more backwards compatible for now

PR #134, opened 2025-01-24
add downloads and lifecycle badges

cjrace:
## Overview of changes

Spotted it was quite easy to add a downloads badge so figured it might be interesting to watch and see if we can grow adoption of this as we add to it.

Added the experimental lifecycle badge while we're still making big changes - expect when we get to v1.0.0 on CRAN we can mark it as stable (thinking either after the next milestone, or after we have done a big update on the layout side).

Adds a ProjectID to the Rproject file, seems to be a default thing R Studio does now so no harm in adding in.

## PR Checklist

## Reviewer instructions

Not a lot to do here! Looks like this
image


sarahmwong (approved):
Nice! Nosey me is very keen to see the downloads badge!

PR #136, opened 2025-02-03
Remove extra button input

sarahmwong:
## Overview of changes

Extra button input header in example app - removed here and fixed a couple of tiny issues at the same time!

Button for Next page was taking you to feedback page - removed two instances of Shiny responding to btn1 clicks so it now takes you to the correct page

button_Input function was called gov_button in the R scripts - function name hasn't changed but just thought it'd be easier to change the name to what the function is called as I found it confusing at first!

## PR Checklist

## Reviewer instructions

Check that the example app looks ok to you now I've removed the extra button input in the third page

Check that the Next Page button on the first page now takes you to second page

Check if you agree with renaming the button_Input file so it matches the function name


inst/example_app/app.R
```diff
@@ -283,7 +283,6 @@ shiny::shinyApp(
               ),
               heading_text("govTabs", size = "s"),
               shinyGovstyle::govTabs("tabsID", data, "tabs"),
-              heading_text("button_Input", size = "s", id = "button_input_tables_tabs_accordions"),
```
cjrace, line 286: One very nitpicky thing, would we add a bit more space between the table and the accordions heading? 
image

cjrace (approved):
Nice - thanks for adding the space!

Issue #70, opened 2024-04-18
Add ability to paginate tables


cjrace:
Paginating big tables is a common ask in dashboards (and is preferrable to internal scrolling within the page). 

It'd be nice to explore if there's something we can add to the existing table function here to allow paginating of tables in an accessible way.

cjrace:
Have a decent version using reactable for this in https://github.com/dfe-analytical-services/apprenticeships-provider-dashboard

cjrace:
We've started a DfE specific version of using reactable with pagination that might be interesting to look at https://github.com/dfe-analytical-services/dfeshiny/pull/87

PR #140, opened 2025-02-07
Sort and paginate tables

### LINKED ISSUE #69: Add ability to sort tables
Sorting tables is a common ask in dashboards, but the standard sort in DT tables tend to be riddled with accessibility issues. It'd be nice to explore if there's something we can add to the existing table function here to allow sorting of columns in an accessible way.

AndyABaker:
## Overview of changes

The previous govTable had no sorting or pagination ability. This new govTable should look just like the old one but with this added functionality, without losing accessibility.

Fixes #69 and #70 

## PR Checklist


R/govTable.R
sarahmwong, line 1: Looking really good and functionality works just as expected, thanks for taking this on Andy! Spotted a small bug I wondered if you had time to address? I think the num_col argument isn't working as expected - instead it takes any column stored as a numeric variable and puts it in the numeric format. E.g. here - if we don't wrap years in speech marks, it will right align it and ignore the specified number cols in the function call, but if we do wrap them in speech marks to force to character, it left aligns:

```
if(interactive()) {

  Months <- rep(c("January", "February", "March", "April", "May"), times = 2)
  Names <- c("Harry", "Sally", "Henry", "Jane", "Emily", "Ant", "Maisie", "Liam", "Kelly", "Alan")
  Years <- c(1995, 1955, 1925, 2001, 1970, 2020, 1960, 1930, 2015, 1990)
  Vans <- c(150, 130, 180, 160, 140, 175, 135, 185, 155, 145)
  Buses <- c(200, 180, 220, 210, 190, 215, 185, 225, 205, 195)
  example_data <- data.frame(Months, Names, Years, Vans, Buses)

  ui <- fluidPage(
    shinyGovstyle::header(
      main_text = "Example",
      secondary_text = "User Examples",
      logo="shinyGovstyle/images/moj_logo.png"),
    shinyGovstyle::banner(
      inputId = "banner", type = "beta", 'This is a new service'),
    shinyGovstyle::gov_layout(size = "two-thirds",
      govTable(
        "tab1", example_data, "Test Table", "l",
        num_col = c("Vans","Buses"),
        col_widths = list(Months = "one-third"),
        page_size = 5
        )
    ),

  shinyGovstyle::footer(full = TRUE)
  )

  server <- function(input, output, session) {}

  shinyApp(ui, server)
}
```

shinyGovstyle.Rproj
sarahmwong, line 1: Please could we revert this? I don't think it's a massive deal just a difference in R versions causing this, but might mean the ID get regenerated by different users >[more info on that here](https://forum.posit.co/t/use-for-projectid-in-rproj-file/196345/2)

tests/testthat/test-govTable.R
sarahmwong, line 1: Once the fix is ready, please can we add a test to check that the fix for the numerical col issue works too?

sarahmwong (changes requested):
Looking great thankyou @AndyABaker! Just one small bug, let me know if you want to chat 😄 

EDIT: Gone back and forth with this thinking it might be best to just do away with the number handling bit entirely as reactable handles this automatically, but I think there are still cases where a user might want to right-align a character column (like if they have commas in large numbers, or pound signs in reporting). Just wanted to capture that requirement here!

R/govTable.R
```diff
@@ -31,91 +31,66 @@
...
+                      page_size = 10) {
 
-  #Create row by row the main bulk of table to insert later
-  main_row_store <- NULL
-  for(i in 1:nrow(df)) {
-    temp_row_store <- createRows(df[i,], num_col)
-    main_row_store <- shiny::tagList(temp_row_store, main_row_store)
-  }
+  # Generate column definitions
+  col_defs <- setNames(lapply(seq_along(names(df)), function(index) {
```
sarahmwong, line 56: my devtools::check() throws up an issue here but I think using `stats::setNames(...` instead works to remove it if you don't mind adding in!

sarahmwong (changes requested):
Thank you again for this Andy, that works fab. Two tiny little fixes if you are ok to action and rerun the devtools::check to update documentation - then happy to approve for you to merge!

sarahmwong (approved):
Perfect thanks Andy!

PR #141, opened 2025-02-17
update to point to correct "for inputId" for selector

sarahmwong:
## Overview of changes

Updates html tag for the sorter label to point to the correct input Id

## PR Checklist

- [ x] I have tested these changes locally using `devtools::check()`

## Reviewer instructions

Check that the label tags look sensible to you now - should match up with what we can see in the gov.uk frontend


cjrace (approved):
All looks great! Nice it ended up such a simple change 😀 

image

Issue #62, opened 2024-04-18
Give examples of the use of text, ordered / unordered lists, and links


cjrace:
Would be helpful to have more examples of how to add text, headers, links within text, and lists of bullet points in a way that takes advantage of the GDS CSS styling. I know there's already the font() function in the app. We found a lot of analysts missed this and weren't doing it correctly in their dashboards.

This could either be just extra examples in the README / test app, or as a gov_text(), gov_list(), gov_link() set of functions for more explicit use.

I have specific examples I can share on this to have a look at from our accessibility audit.

fingertipsy:
Could you maybe share some examples here until this is picked up in the documentation? It would be so helpful! I have been struggling to figure out how to include correctly styled links and lists within an accordion

cjrace:
Certainly @fingertipsy! Sorry I'd missed this, there were a few different places issues around this were picked up. Though here's an example of how currently I've formatted text in R Shiny by having a custom CSS stylesheet (not necessarily the best way, this is just how we've worked it for now!).

1. Create or edit a CSS file in your project folder - e.g.  `www/dfe_shiny_gov_style.css`

Within that include something like this (add or remove sections and edit settings as you need)

```
/* Fonts ------------------------------------------------------------------- */
/* main body text */
body {
  font-family: "Helvetica Neue", "Arial", sans-serif;
  font-size: 19px;
  margin-bottom: 20px;
}

p {
  font-family: "Helvetica Neue", "Arial", sans-serif;
  font-size: 19px;
  margin-bottom: 20px;
}

.h1,
h1 {
  font-family: "Helvetica Neue", "Arial", sans-serif;
  font-size: 48px;
  font-weight: 700;
  line-height: 1.0416666667;
  margin-top: 0;
  margin-bottom: 30px;
}

.h2,
h2 {
  font-family: "Helvetica Neue", "Arial", sans-serif;
  font-size: 24px;
  font-weight: 700;
  line-height: 1.0416666667;
  margin-top: 20px;
  margin-bottom: 20px;
}

.h3,
h3 {
  font-family: "Helvetica Neue", "Arial", sans-serif;
  font-size: 20px;
  font-weight: 700;
  line-height: 1.0416666667;
  margin-top: 20px;
  margin-bottom: 20px;
}

/* header text */
.govuk-header__link--service-name {
  font-family: "Helvetica Neue", "Arial", sans-serif;
  font-size: 30px;
}

/* feedback banner */
[...]

PR #142, opened 2025-02-21
create gov_list function

### LINKED ISSUE #123: Accordions only open when you click "hide" or "show"

jenisayshi:
## Overview of changes

Create gov_list function and added to run_Example

Linked to issue #62 

## PR Checklist

- [ ] I have tested these changes locally using `devtools::check()`


R/govLayouts.R
sarahmwong, line 1: could this sit as a separate function in a new file instead? Know gov_text is in here at the moment, but I think it's not an obvious place to look for gov_list/could get lost in general layout stuff that we are probably going to faff with in the next milestone with widescreen options!

shinyGovstyle.Rproj
sarahmwong, line 1: Please can you revert this? I think its just from running an older version of R, we should probs stick in the gitignore file actually

sarahmwong (changes requested):
Thank you Jeni this looks great! Suggested a couple of small changes to how it sits but otherwise happy for you to tick the boxes and merge. @cjrace was there anything on #62 that you think remains outstanding now? I think the external link thing is covered, there is a gov_text() function that already exists here too

cjrace:
> Thank you Jeni this looks great! Suggested a couple of small changes to how it sits but otherwise happy for you to tick the boxes and merge. @cjrace was there anything on #62 that you think remains outstanding now? I think the external link thing is covered, there is a gov_text() function that already exists here too

I think this covers it pretty much, happy for this PR to be linked so that it closes that issue @sarahmwong / @jenisayshi

shinyGovstyle.Rproj
sarahmwong: yes please!

sarahmwong: NB - we discussed this morning and you will look to update R instead of adding to git ignore!

R/govList.R
```diff
@@ -0,0 +1,54 @@
+#' @rdname layouts
+#' @param list vector of list
+#' @param style options: "none", "bullet", "number". defaults to "none".
+#' @export
```
sarahmwong, line 4: Thanks Jeni! Please can you add in an example to the govList R file like how the other functions are set up? Then running devtools::check() should update the documentation, should be good to go after that though!

Issue #129, opened 2025-01-07
Move away from putting all examples in README and point to showcase app / rest of documentation


cjrace:
## Describe the solution you'd like
We currently have examples of the code in 3 places:

1. Function documentation on https://dfe-analytical-services.github.io/shinyGovstyle/ and in R Studio help panel
2. `run_example()` / showcase app at https://department-for-education.shinyapps.io/shinygovstyle-example-app/ (or in the console)
3. In the README

This is starting to feel quite duplicative, meaning we need to maintain 3 sets of examples for the functions. My suggestion would be to drop the README examples and instead lean on 1) and 2) more now we have the live URLs people can easily browse.

What we could do is record a GIF of the showcase app in action as a teaser in the README to encourage more people to view it, that then wouldn't need maintaining as much.

## Describe alternatives you've considered
Both 1. and 2. have additional value over the README. 1 is standard practice for R packages and a base requirement, 2. gives the ability to see the components in action and makes it easier for us to test, so I don't think it's worth dropping either of them instead.

We could keep a minimal set of examples in the README, though that should be limited to something that gives an overview of the package rather than every function.

## Additional context
First raised in #128 in discussion with @rmbielby.

Interested to know others thoughts, particularly @sarahmwong?

sarahmwong:
really good point thank you both for raising! Linking to the function documentation and example app is a great idea - the pace of changes of styling mean keeping on top of README PNGs could become painful very quickly! This relates to #76 which I'm happy to close to focus on this instead?

cjrace:
Yeah good spot @sarahmwong - I agree we close #76 as not planned as there's less use for updating screenshots when we always have the live link and in this work we'll remove all the screenshots anyway!

PR #143, opened 2025-03-03
split old and new gov table functions out 

sarahmwong:
Splits old and new gov table functions out so that any code written with the old table does not fall over. Small fix to the way captions are dealt with in the new interactive function too.

## Overview of changes

Previous commit to main broke old code (sorry I totally forgot about this!) - splitting govTable (old function) and govTable_interactive (new function) into two completely separate options. I think there is still a genuine need for both in hindsight, so good to keep the old in! govTable() works well for small feature tables, whereas govTable_interactive() lets users filter through more rows and cols of data.

## PR Checklist

## Reviewer instructions

- Check that the old function still works as expected. Looks fine in run example and passes reinstated tests
- Check you're happy with how we are dealing with the caption for govTable_interactive() - previously was squishing up when we used a long title, switched approach to match govTabs and this looks more like it!


sarahmwong:
left the README as is as we are going to stop updating this as part of this milestone #129

R/govTable_interactive.R
```diff
@@ -0,0 +1,99 @@
...
+#'
+#'   ui <- fluidPage(
+#'     shinyGovstyle::header(
+#'       main_text = "Example",
+#'       secondary_text = "User Examples",
+#'       logo="shinyGovstyle/images/moj_logo.png"),
+#'     shinyGovstyle::banner(
+#'       inputId = "banner", type = "beta", 'This is a new service'),
+#'     shinyGovstyle::gov_layout(size = "two-thirds",
+#'       govTable(
```
cjrace, line 34:
```suggestion
#'       govTable_interactive(
```

Think this needs to the interactive version?

cjrace (changes requested):
All tests fine, and no longer breaks the place where we spotted the breaking change. Just one small copy / paste error in the examples.

Agree having both makes sense, as you say, thinking about it there's good reasons for the simple variation too!

I think there might be some additional custom styling we can add (focus styles and reordering headers) that I've done elsewhere in DfE though happy for this to go in as is for now and we can look at that at a later point

cjrace (approved):
Thanks! Happy for this to go in

PR #145, opened 2025-03-11
add gov_list function

jenisayshi:
## Overview of changes

Add gov_list function.

## PR Checklist

## Reviewer instructions

Repeat of same pull request with new branch off main to stop git conflicts in example_app


R/govList.R
```diff
@@ -0,0 +1,82 @@
+#' @rdname layouts
```
sarahmwong, line 1: think this is forcing roxygen to pop this in the layouts documentation - if you rename to gov_list does that create a new.Rd file for this?

sarahmwong, line 1: please could we also change the .R file name to gov_list so it matches the function name? 😄

sarahmwong (changes requested):
Thanks for sorting the messy conflict! Looks like it has sorted it here. Some teeny documentation fixes if you're ok to action :)

R/govList.R
jenisayshi: both bits of doc changes should be fixed now :)

sarahmwong (approved):
thanks jeni!!

sarahmwong:
closes #62

Issue #93, opened 2024-07-19
Update file upload to be consistent with GDS styling


sarahmwong:
Picked up when looking at updating the styling to latest GDS v 5.4 

The reason that it's not picking up the GDS styling is that shiny implements it's own tailored upload input that's different from the standard html method that the GDS assumes. 

See further details in #91 unmerged and closed PR

PR #148, opened 2025-03-17
update to v.5.9.0 frontend. 

sarahmwong:
## Overview of changes

Update to latest front end version

Update css manual changes file to remove the two png sizes, just svg for govuk crest now

Looks like they've added a load of file upload stuff which should help with #93 in the next milestone

## PR Checklist

## Reviewer instructions

Checked with run_example() and dont think this introduced any breaking changes - please can you check this is the same for you?

Existing PR open to fix value boxes shouldn't affect these if we want to merge this first - no CSS changes in that PR (though that PR is also ready to merge!)


cjrace (approved):
Checked it running locally side by side with the live site and all looks good, didn't see anything concerning in the changes either so hopefully all pretty smooth and easy!

PR #152, opened 2025-03-26
External link example added to footer

### LINKED ISSUE #138: Adapt the footer function to also take external links
We found an example of wanting an external link in the footer as well as internal links, in https://github.com/dfe-analytical-services/shiny-template/pull/120#pullrequestreview-2590421956

Ideally we'd allow external links as well as internal links, and where they're used they make use of our `external_link()` function, but with the opens in new tab warning there for screen readers only (as I've not seen any GOV site have opens in new tab in a footer link, so I'm assuming it's expected behaviour that it changes the page given the link context in the footer).

Not considered any other alternatives, assuming we can find a way to pass in both external and internal links, even if we move to accepting a dataframe as the input so you can customise the link text / link url / whether it is external

## Additional context
Example GOV.UK site with a mix of links in the footer - https://explore-education-statistics.service.gov.uk/

alex-rogers-hub:
## Overview of changes

External link in footer example included here, as in shiny_template

Fixes #138 

## PR Checklist

- [ ] I have tested these changes locally using `devtools::check()`


cjrace:
Thanks @alex-rogers-hub! Do you think it would be possible to include the code / adapt the existing function within the package code itself so that users of the package can just use the `footer()` function to get this behaviour?

alex-rogers-hub:
I think a secondary function to add to the server.R code would be best here - an observeEvent section is currently needed in server.R for each link, instead of what currently goes inside I can place a function in there which simplifies things

cjrace:
> I think a secondary function to add to the server.R code would be best here - an observeEvent section is currently needed in server.R for each link, instead of what currently goes inside I can place a function in there which simplifies things

Hey @alex-rogers-hub, thanks! I agree that a separate server function makes sense for the code as you have it, though it's just made me wonder if there's another way we can do this without needing a server function / modal popping up? Maybe there's some way to pass in a named list of links into the function and then recognise if something is an external link or an internal link to then generate the needed HTML without any need for a server call / observe event?

I could image passing a list like this into the function in the UI and then not needing a server as we could then generate the right HTML from this alone?

```
footer(
links = list(
 "Accessibility" = "internal",
 "Support" = "internal",
 "Privacy notice" = "actual.url.of.privacy.notice"
)
```

Not tested or tried it, and I am happy to fall back to the code as we have it if not, but figured it's worth exploring all options before we commit the package one way or the other!

alex-rogers-hub:
@cjrace looks like a possible solution, I can certainly give it a go. Couple of considerations:
 - "internal" currently assumes tabset - existing setup allows for relatively simple additions if you want to go outside of tabs/external. Could potentially do something similar with the list input having something like: "Accessibility" = "internal tabs"
 - I'll look at coding it this way but can't promise it will be ready and tested for Monday

R/footer.R
```diff
@@ -200,3 +229,53 @@ footer <- function(full = FALSE, links = NULL) {
...
+      warning("Link name provided is NULL for ", link)
+      link_name = link
+    }
+    shiny::tags$li(
+      class = "govuk-footer__inline-list-item",
+      external_link(
+        link,
+        link_name,
+        add_warning = FALSE
+      )
```
cjrace, line 266: Using external link overwrites the CSS classes (or at least adds more to it), which means it's blue rather than the expected black (e.g. - https://explore-education-statistics.service.gov.uk/)

image

cjrace (changes requested):
Nice - this is even neater than the solution I had in mind!

Few things (as well as one inline comment):

1. Can we have more tests on this - making sure we cover mixes of external links and internal links, the validation around names being included etc
2. The spacing ideally would line up with the left hand side of the OGL line in the footer when using `full=TRUE`
image

EES example
image

3. Potentially related to 2., in the example dashboard the internal 'cookies' link is no longer appearing but should be?

Footer before
image

Footer in this branch
image

Code
```
shinyGovstyle::footer(
      TRUE,
      links = c(
        "Cookies",
        `GitHub repository`="https://github.com/dfe-analytical-services/shinyGovstyle")
      )
```

4. Can you update the docs for the footer function to reflect these changes? (mostly thinking description / param description)

rmbielby:
Made some updates along the lines of the suggestions, so...
- added some extra tests, 
- fixed the bug around the mixed internal / external not working (I've put a fix so it would have worked, but also switched up the example to use the more explicit form as well)
- added the ability to pass a specific class to external_link and wired up the footer to pass it's preferred link styling
- some slight updates to the docs in there as well
- removed the is.vector() validation given that pretty much anything an analyst would be likely to provide (i.e. vectors, single string, list) would actually pass the test (and the function works with them)

rmbielby:
> Thanks @rmbielby - tested these and it seems to all be working as expected, the extra test cases and updates to the documentation are helpful.
> 
> One question though, what is the use case for someone using a different CSS class in this function?
> 
> Given we're trying to copy the GOV.UK approach I'd be tempted to just lock them into the GOV.UK link styling (so instead of having it as an argument with a default, remove the argument and leave the default as the way it is)

The footer _is_ the use case for the extra class in the external_link function. If we want to lock it down, the two options that spring to mind would be:
- remove `class` as a param from `external_link()` and replace it with `footer`, such that the footer class get's applied when `footer=TRUE`
- add validation to class such that it can only be valid gds link classes (i.e. if there are any other link classes outside of standard and the footer).

cjrace:
Fair point! I think the first option of make it an optional 'footer=TRUE' argument would be the clearest and least likely to be accidentally misused

PR #155, opened 2025-06-28
Use snapshot testing for markup that wraps actionButton()/actionLink()

cpsievert:
Hello, I'm a core contributor to Shiny, and this PR is in anticipation for https://github.com/rstudio/shiny/pull/4249, where we plan on making changes to the return value of `actionButton()`/`actionLink()`, which will break these tests.

Generally speaking, tests shouldn't be making such strong assumptions about the return value of Shiny UI functions. This is especially true of tests running on CRAN -- it blocks us from doing a release for changes that have a small chance of breaking existing behavior.

That said, you may have CSS/JS that truly does rely on a certain HTML markup produced by Shiny. In that case, you should consider using [snapshot testing](https://testthat.r-lib.org/articles/snapshotting.html) to capture and maintain relevant markup expectations. Crucially, since snapshots are inherently brittle, they do not run on CRAN, but will run everywhere else, and can be easily updated when snapshots change.

### DIFF
--- a/tests/testthat/_snaps/footer.md
+++ b/tests/testthat/_snaps/footer.md
@@ -1,0 +1,127 @@
+# footer links add correctly
+
+    Code
+      footer_with_links
+    Output
+      <footer class="govuk-footer " role="contentinfo">
+        <div class="govuk-width-container ">
+          <div class="govuk-footer__meta">
+            <div class="govuk-footer__meta-item govuk-footer__meta-item--grow">
+              <div>
+                <h2 class="govuk-visually-hidden">Support links</h2>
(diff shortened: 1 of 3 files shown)


cjrace:
Thank you for this @cpsievert - I have made a couple of minor tweaks before I'll merge in (adding our R-CMD-check GitHub Action to a regular schedule, activating the 3rd edition of testthat across the whole package, and adding a snapshot on the backlink test).

I agree that this should prompt a wider review of our testing - have raised #156 covering this for us to pick up before our next release.

cjrace (approved):

PR #166, opened 2025-07-16
157 fix css on govreactable

sarahmwong:
## Overview of changes

Small tweaks to where the div points to. Couldn't get this working in the separate css file but I think fine to keep as an addition to manual css as it's fairly small!

## PR Checklist

## Reviewer instructions

Check that the yellow highlight on col header, row and pagination works fine for you

NB- are we wedded to the lines between rows like in govtable if we've got this hover? I don't really like the width of the rows here but if we remove `class = gov-uk-table__cell` then it looks a bit more like the static table in terms of spacing:

image


R/govReactable.R
```diff
@@ -69,7 +69,7 @@ govReactable <- function(
       reactable::colDef(
         name = col,
         sortable = TRUE,
-        headerClass = "govuk-table__header",
+        headerClass = "bar-sort-header",
```
cjrace, line 72: 🤦

cjrace (commented):
Thanks @sarahmwong - can't believe I spent as long as I did failing to get this working now!

1. Styling works for me
2. I don't know if we're wedded to the lines between rows, though the GOV.UK table component does have them so we probably should have them too?
3. I think I prefer the widths without the gov-uk-table__cell class (look at the Vans header, seems to sit weirdly when the class is applied)

Without gov-uk-table__cell 

Image

With gov-uk-table__cell

Image

PR #170, opened 2025-07-17
Fix example app reactive table outputs

### LINKED ISSUE #169: Published example app reactive table outputs are incorrect
## Overview
Works fine locally, but on [the example app
](https://department-for-education.shinyapps.io/shinygovstyle-example-app/), the second reactive reactable table has the wrong data populated

## Expected behaviour
Expect a cut of the data filtered red/blue as in the static table above

## Screenshots

Image

## Additional context
Pretty sure this was happening before the latest commit to main (before the css fix) as I just checked the example app before merging!

cjrace:
## Overview of changes

@sarahmwong I've figured this out and this PR fixes #169.

It was the same PR where I brought the base R pipe in (#168) but it wasn't the pipe, it was removing the dplyr library call that then changed filter to be the [base R filter](https://www.rdocumentation.org/packages/stats/versions/3.6.2/topics/filter) instead (from the stats package), which behaves quite differently!

I've swapped it for `subset()`, which is the closest base R equivalent to `dplyr::filter()` and all seems to be working fine again.

Sorry about this, when I checked on that branch I must have already had dplyr loaded in the environment without realising!

## PR Checklist

## Reviewer instructions

Run the example app and check it's working for you locally.

### DIFF
--- a/inst/example_app/app.R
+++ b/inst/example_app/app.R
@@ -705,7 +705,7 @@ shiny::shinyApp(
     )
 
     filtered_data <- reactive({
-      example_data[example_data$Colours == input$colourFilter, ]
+      subset(example_data, Colours == input$colourFilter)
     })
 
     output$interactive_table_test <- renderGovReactable({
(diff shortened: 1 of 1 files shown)


sarahmwong (approved):
Ooooh great spot - really weird and must have been the same here keeping the same R session open and testing, have learned for next time!!

PR #171, opened 2025-07-17
GDS-like file upload styling

rmbielby:
## Overview of changes

I've combined existing css classes and some custom styling to force something along the lines of the GDS file upload input. I'm assuming we're going to want to fold the custom styling into css modifications, but I'm just doing it this way for now for a review and in case actually we're fine with the explicit styling within the function.

## PR Checklist

### DIFF
--- a/R/file_Input.R
+++ b/R/file_Input.R
@@ -119,14 +119,19 @@ file_Input <- function(
 
     shiny::div(
       id = paste0(inputId, "file_div"),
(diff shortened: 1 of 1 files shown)


rmbielby:
Thanks for sorting those bits. Looks good here. I can't approve, so leaving to you / @sarahmwong to finish up the approval & merge.

cjrace (approved):

Issue #164, opened 2025-07-14
Test dependabot alerts


sarahmwong:
## Overview
Once #161 is merged into main, test the dependabot alerts for gov.uk

## How to reproduce
Merge to main - dependabot should figure out that we're out of sync with the latest frontend package

## Expected behaviour
Dependabot alert should be raised in contributors inbox when the frontend is out of date with the latest version available (as it currently is in main/master)

PR #175, opened 2025-09-23
Fix lint issues

### LINKED ISSUE #162: Resolve lintr issues and enable checking moving forwards
Address lintr issues and reenable the workflow.

Once reenabled we should then enforce it on all pull requests like our other workflows.

We could avoid doing this, as it's not urgent or blocking us doing CRAN releases etc. Though I think it's worthwhile getting to the point where we can turn on lintr so we can maintain tighter standards and conventions moving forwards.

Most (though not all) of the issues are related to variable and argument names not following the snake_case convention, so for those we should do one of the following:

1. Accept breaking changes, and warn users that function and argument names will change (risky given it's already in use across departments)
2. Add new alternative variable names and arguments and advise users of deprecation of older names in future versions (potentially more work than it's worth and messy)
3. Add nolint tags to affected lines (I'd recommend as the easiest for now)

Then also fix / address any other warnings.

## Additional context
This will be a bit of a fiddly one.

cjrace:
## Overview of changes

Bit of a monster here, though ultimate goal was to enable the lintr workflow and resolve #162.

- Consciously used `# nolint` in places like lint issues around existing function names instead of fixing, as fixing would create breaking changes for users
- Updates the lintr workflow file so it will run more often than once in a blue moon 🌑 
- Switched to roxygen2md (was prompted to by one of the packages)
- Removed magrittr pipe as we can use base R pipe now, removes a dependency
- Enables PRs to be raised by dependabot (to test #164 and see if it will pick up on the newer version of govuk-frontend on npm through a PR as there's no alerts currently when we'd expect to see some)
- Switched a few examples to only use `if (interactive())` at the end, which is shorter in lines, and also lets us find out if the examples error - found at least one example where there was a typo that got flagged, so managed to fix that!
- A few other minor tweaks to consistency in docs / code, none of which should affect the functionality of the package

## PR Checklist

## Reviewer instructions

Too many changes to check manually probably... Make sure when viewing files changed you have whitespace turned off.

- Check you can run the tests on the branch locally
- Run and check the showcase example app using `run_example()` and compare to the current main version, to make sure I've not messed any of the components up - https://department-for-education.shinyapps.io/shinygovstyle-example-app/
- Scan as much as you have the stomach for to see if the changes make sense!


DESCRIPTION
```diff
@@ -1,39 +1,39 @@
...
+Package: shinyGovstyle
+Title: Custom Gov Style Inputs for Shiny
+Version: 0.1.0.9000
+Authors@R: c(
+    person("Ross", "Wyatt", , "ross.wyatt@justice.gov.uk", role = c("aut", "cre")),
+    person("Cameron", "Race", , "cameron.race@education.gov.uk", role = "ctb"),
+    person("Sarah", "Wong", role = "ctb"),
+    person("Richard", "Bielby", , "richard.bielby@education.gov.uk", role = "ctb"),
+    person("Charlotte", "Foster", , "charlotte.foster@education.gov.uk", role = "ctb"),
+    person("Jeni", "Martin", role = "ctb")
```
sarahmwong, line 10: ooo reminded me can we add Andy Baker in for his work on the reactables and value boxes please? Andrew.Baker1@dhsc.gov.uk :)

sarahmwong (changes requested):
Woooft thank you for sorting this Cam - sifting through all the #nolint variables to mark must have taken an age! All looks good to me, checks run fine and tested against the example app. Had a look at 15 or so of the fixed R files too and looks good to me. Just one small change to request to add Andy into contributing or else I'll definitely forget before the next CRAN update!

DESCRIPTION
cjrace: Of course! Added 😄

PR #203, opened 2025-11-06
docs: use families, update pkgdown reference

### LINKED ISSUE #200: [Feature]: Update documentation site to group related functions
### Description

We should explicitly organise the reference list within the docs site using the _pkgdown.yml. E.g. https://dfe-analytical-services.github.io/eesyscreener/reference/index.html

### Use case

As a user, I want to easily glance at available functions, so that I can understand everything that is available within the package with minimal effort.

### Why this is important

Helps us to show where functions do similar roles (e.g. gov_layout v gov_main_layout), as well as highlight things like exported data and related groups of functions.

### Additional Context

Should copy the layout of the showcase app for most individual components so it's easily / familiar for users looking at both the documentation and the showcase.

jack-davison:
## Overview of changes

Fixes #200.

I can tell that shinyGovstyle is going to be really useful, but I struggled to search through the available functionality.

This PR adds:

- roxygen2 function families, which cross-reference R documentation.

- an ordered pkgdown function reference page.

The order is the same as the showcase website, as outlined in #200.

Also updated NEWS.md.

## PR Checklist

- [X] I have updated the documentation (if needed)
- [X] I have added or updated tests for these changes (if applicable)
- [X] I have tested these changes locally using `devtools::check()`

## Reviewer instructions

There is an "Other components" section I don't like very much - potentially some decisions to be made by package authors to classify these functions further.

image

### DIFF
--- a/tests/testthat/_snaps/backlink_Input.md
+++ b/tests/testthat/_snaps/backlink_Input.md
@@ -3,5 +3,7 @@
     Code
       backlink_Input("backId")
     Output
-      <a class="action-button govuk-back-link" href="#" id="backId">Back</a>
+      <a class="action-button action-link govuk-back-link" href="#" id="backId">
+        <span class="action-label">Back</span>
+      </a>
 
--- a/tests/testthat/_snaps/contents_links.md
+++ b/tests/testthat/_snaps/contents_links.md
@@ -4,7 +4,9 @@
       contents_check
     Output
       <div class="govuk-contents">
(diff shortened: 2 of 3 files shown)


cjrace:
Hey @jack-davison, sorry for the delay in reviewing this! Really appreciate you taking a look and putting the time to raising this PR. Looks fantastic!

I've made a commit on this branch with suggestions of families we could use for the 'other' list you had - let me know what you think!

@sarahmwong - worth you doing a quick check too, and if we're all happy then we can get this merged in, will make a big difference for users of the package to understand and navigate the available functions 😀

PR #207, opened 2025-12-18
Apply snapshot update to backlink test

cjrace:
## Overview of changes

We've had errors flagged on CRAN - https://cran.r-project.org/web/checks/check_results_shinyGovstyle.html. Effectively the backlink test was no longer working across a selection of OS's.

image

image

We'd already moved to a snapshot approach, but for speed of getting this fixed before the deadline CRAN have set, I've branched off from the v0.1.0 tag (last version we sent to CRAN), to make a v0.1.1 patch that we can bundle and submit to CRAN and then re-integrate with the existing development changes on main, which we can look to 

## PR Checklist

## Reviewer instructions

No review needed, will just let the actions run and assuming it passes on all OS's then I will just bundle from here and submit to CRAN as we've already approved these changes to main in , though tagging @sarahmwong for awareness as we'll need to agree the best approach for representing this version as a tag in the repo (maybe merging all outstanding PRs and then forcibly rebasing main on top of this branch so that v0.1.1 appears neatly in the commit history?)


cjrace:
This was an isolated fix that I submitted to CRAN and is now 89 commits behind main - given that rebasing all of main on top of this (the technically correct linear history) would then break everyone elses clones of the repo and be disruptive, I'm instead just going to merge this in on top of the changes. Technically the Git history won't be in the right order, though CRAN doesn't care about that (it just gets the bundle from a point in time).

Seems like the best option and makes it easier and less error prone too (fewer conflicts to sift through)

Issue #208, opened 2026-01-09
[Bug]: Contents pane removes word wrap on scroll


sarahmwong:
### Overview

Identified when fixing #194 - word wrap on the subcontents spills over when a user scrolls down. Would expect the wrap to stay in place on scroll, otherwise risks spilling into the main body of the app

### Steps to Reproduce

Inspect showcase app, scroll down.

subcontents on page open (expected behaviour):
<img width="1544" height="1038" alt="Image" src="https://github.com/user-attachments/assets/2c5f6a9f-b310-4242-a25a-1323c0d09e3f" />

subcontents on scroll:
<img width="1544" height="1038" alt="Image" src="https://github.com/user-attachments/assets/853899fd-b2af-40fd-a641-69e793e563f1" />

### Expected Behavior

Subcontents links remain wrapped on scroll, as in first screenshot above

### Additional Information

Think there is something in the contents_link.js and getting it to respond to certain scroll behaviour to fix in this script

JEdholm1:
Hello, I'm also finding this issue. I saw that col-widths-lg is set as 2,9, so doesn't add up to 12 and wondered if that might be causing it?


<img width="1820" height="759" alt="Image" src="https://github.com/user-attachments/assets/0f5a0dd8-7c35-4630-94f1-f2e6a56e91d9" />

cjrace:
Closing this as since v0.2.0 (and discussion in #176), we've now deprecated the contents links functions in favour of the horizontal nav.

Issue #201, opened 2025-10-13
[Bug]: Remove `full_width_overrides()`


cjrace:
### Overview

I created it, but knew even then it shouldn't exist. We should update the core CSS to include the adapted full width CSS and add into any related functions.

### Steps to Reproduce

Not applicable.

### Expected Behavior

It should be default styling within the other components as to whether you have full or partial width, should not need an additional function for it.

PR #209, opened 2026-01-09
Update rems and weights for links

### LINKED ISSUE #194: [Bug]: Link text in contents is too small for the colour contrast
### Overview

Some of the text in the showcase app is too small for the colour contrast (to meet AAA).

Image

Image

### Steps to Reproduce

Inspect the showcase app.

### Expected Behavior

For sites to meet AAA, they must comply with WCAG 2.2 checkpoint 1.4.6 for colour 
contrast. As far as contrast ratio is concerned, it must at least be:

- if text is not bold and its size is less than 18pt/24px/1.5em/150%: 7:1 for AAA level;
- if text is not bold and its size is at least 18pt/24px/1.5em/150%: 4.5:1 for AAA level;
- if text is bold and its size is less than 14pt/19px/1.2em/118%: 7:1 for AAA level;
- if text is bold and its size is at least 14pt/19px/1.2em/118%: 4.5:1 for AAA level

The GOV.UK link contrast is around 5-6:1, so we need to ensure the text is at least 1.5em (or 1.2 em and bold)

### Additional Information

Page 44 of the September 2025 accessibility audit report.

sarahmwong:
## Overview of changes

Increases rem on contents links to 1.2 to meet WCAG AAA as outlined in #194 and adds bold weight. Updated for external and download links too.

Identifies new bug #208 where contents pane spills over onto main app - faffed with the interim fix on full_width_overrides() just so the example app doesn't look weird! 

## PR Checklist

## Reviewer instructions

Check that links now meet the WCAG standards but function still remains in example app

### DIFF
--- a/inst/www/css/govuk-frontend-5.9.0.min.css
+++ b/inst/www/css/govuk-frontend-5.9.0.min.css
@@ -7968,3 +7968,13 @@
     box-shadow: inset 0 -0.188rem 0 0 #1d70b8;
 }
 
+/* download links--------------------------------- */
+
+.shiny-download-link {
+    font-size: 1.2rem;
+    font-weight:bold;
+}
+
+
+
+
--- a/inst/www/css/govuk-frontend-5.9.0.min.css
+++ b/inst/www/css/govuk-frontend-5.9.0.min.css
@@ -14,7 +14,9 @@
     -moz-osx-font-smoothing: grayscale;
     text-decoration: underline;
     text-decoration-thickness: max(1px, .0625rem);
-    text-underline-offset: .1578em
+    text-underline-offset: .1578em;
+    font-size: 1.2rem;
+    font-weight: bold;
 }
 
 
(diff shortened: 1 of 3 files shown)


inst/www/css/govuk-frontend-5.9.0.min.css
```diff
@@ -14,13 +14,15 @@
     -moz-osx-font-smoothing: grayscale;
     text-decoration: underline;
     text-decoration-thickness: max(1px, .0625rem);
-    text-underline-offset: .1578em
+    text-underline-offset: .1578em;
```
cjrace: Should these changes be in `css_changes.md` too?

cjrace (changes requested):
Thanks @sarahmwong! A couple of thoughts:

1) The cookies link doesn't seem to be caught by these changes?

Image

2) Increasing the link text size makes it out of line with the normal text size, it feels a little jarring? Especially when you look at the beta banner - would it be possible to move all of the text to the same size - how does that look?

3) I'm guessing the size for detail components might need to be increased too?

sarahmwong:
Going to start a fresh PR for this with all the new css changes in 6.0.0 think it makes sense to, also #201 negates the need to fix the issue #208

PR #216, opened 2026-03-30
Fix missing programmatic names for radio and checkbox options

### LINKED ISSUE #189: [Bug]: Missing programmatic names for radio and checkbox options
### Overview

The individual radio and checkbox options aren't marked up with clear names for assistive tech. They have visual labels but these aren't clear to assistive tech.

### Steps to Reproduce

Inspect the HTML from the example app for radio buttons, download radios and checkboxes.

### Expected Behavior

Screen reader comments:

> “While navigating the page with JAWS from the ‘JAWS forms list,’ I found that users will only hear the announcement of ‘check box’ at the time of navigating. Choosing to navigate in a different way such as the ‘arrow keys’ for example, causes JAWS to provide the function of each check box. Although this is a work around, I would expect the check box items to announce a clear indication to the function and related content, such as ‘which types of waste do you transport waste from animal carcasses. Including a clear indication to the function of the content, will resolve this issue for users of screen reading software.”

Dragon NaturallySpeaking comments:

> “The radio buttons and check boxes do not respond to their visual titles. I am unable to access the radio buttons or check boxes when I use their visual titles with voice activation. For example, ‘click yes or click waste from animal’ Dragon does not respond and says ‘please say that again’ this extremely frustrating. Because of this functionality I have to switch to standard commands to be able to activate the components, to do this I will use the command click button and/or click box. This takes longer to do as there are multiple components to access. When I say the command click button Dragon will tag all four radio buttons at once, I will have to choose the corresponding tag to activate the radio button I want to activate. I would expect all actionable components to have a visual label, as they all have a label directly to the right to them, I would expect to be able to activate the component with this. when I use the command click yes, I would expect Dragon to tag both yes radio buttons. Then when I use the command click farm I would expect the check box to automatically activate. This makes it more accessible to access with Dragon commands as it minimises the steps to activate them and makes it more intuitive.”

### Additional Information

Solution: interactive elements must have a programmatic accessible name to ensure that the user can understand the purpose of the element or interact with the element as expected. In this case, each interactive element has a visible label, and the visible label must be programmatically associated to the relevant input by using the for and id attributes.

For example, currently
```
<input type="radio" name="name_changed" value="Yes" class="govuk-radios__input">

<input type="radio" name="name_changed_stacked" value="Yes" class="govuk-radios__input">

<input class="govuk-checkboxes__input shiny-bound-input" id="op1" name="checkID" type="checkbox" value="op1">
```

Suggested
```
<label for="firstExample" class="govuk-label govuk-radios__label">Yes</label>

<input type="radio" name="name_changed" value="Yes" class="govuk-radios__input" id="firstExample">
```

Page 28 of the September 2025 accessibility report

cjrace:
## Summary
- Adds `id` attributes to individual radio button `<input>` elements (format: `{inputId}-{index}`) and corresponding `for` attributes to their `<label>` elements
- Adds `for` attributes to checkbox `<label>` elements, linking them to the existing `id` on each checkbox `<input>`
- Adds tests verifying `id`/`for` association for both components

Closes #189

## Test plan
- [x] Run the example app (`inst/example_app/app.R`) and inspect the HTML for radio buttons and checkboxes — each `<input>` should have a unique `id` and each `<label>` should have a matching `for` attribute
- [x] Test with a screen reader (e.g. NVDA) — each radio/checkbox should announce its label text, not just "radio button" or "check box"


cjrace:
Tested this with NVDA locally and all seems to be working as detailed / expected now 🥳

PR #220, opened 2026-04-09
Fix header wrapping on small screens (and prep for next CRAN release)

cjrace:
EDIT: A lot of the files changed were showing every line having changed the line ending, I think this is us having different local Git settings so I've added a project level .gitattributes file and normalised all of the files in the repo, so while it won't help us on this PR it should prevent that happening in future at least!

# Overview of changes                                                                                                                                                                                                                                                                                                                                                                                                                               

Managed to fix a bug with the header() component on small screens wrapping awkwardly, and then got carried away and bundled in all remaining prep work needed to submit v0.2.0 to CRAN.

92 changed files is a bit misleading, the core fixes are in R/header.R and inst/www/css/header-overrides.css. Beyond that the majority of affected files are related to my renaming of the header arguments (triggered by a conversation I had with Rich the other day).                                                                     

## Header fix (original scope):
- Fixed sizing and alignment of the header on small screens
- Fixed service_name text wrapping on narrow viewports
- Added a new inst/www/css/header-overrides.css to handle these layout adjustments (felt it made sense to given that GOV.UK and us are now diverging on the header component, so we're not editing theirs anymore and instead are effectively maintaining our own)

## Further header improvements:
- Renamed main_text → org_name and secondary_text → service_name for clarity (wanted to do this for a while and figure this might be the best point to do so!)
- Old argument names are deprecated with lifecycle warnings pointing to the new names

## CRAN prep:
- Updated DESCRIPTION to tidy format and CRAN-ready metadata
- Resolved CRAN check issues and notes
- Updated NEWS.md for v0.2.0 release
- Updated README.md and replaced the demo GIF (feel free to recreate this / update it if you want it to show other parts of the example app off!)
- Updated vignettes

All prep is aiming at v0.2.0 release (figure that we can then immediately start preparing fixes in for v1.0.0 in a couple of months time.

PR Checklist
- [x] I have tested these changes locally using devtools::check()

## Reviewer instructions

The original fix is in R/header.R and inst/www/css/header-overrides.css. Check that is working as expected first in the example app.

The rest of the changes are CRAN prep — widespread but mostly mechanical (roxygen doc regeneration, DESCRIPTION tidying, removing stray @keywords tags that were causing issues for the PDF manuals). The main things worth reviewing are:
- The deprecated argument handling in header() — does the lifecycle pattern look right?
- NEWS.md — does the v0.2.0 changelog accurately reflect what's changed since v0.1.0? (it is a lot of things!!)


R/header.R
sarahmwong, line 85: should update all these too "with" as above for consistency - much neater!

sarahmwong: replace with lifecycle::is_present for consistency

sarahmwong (changes requested):
Looks great Cam really nice fix, works on mobile and desktop. Just requesting these small changes to the way we code in the other deprecate warnings in header.R just bring all in line with how you've structured them here (much neater/more robust than my earlier attempt!)

R/header.R
cjrace: Don't think these need a 'with' as we're removing them, so there's no replacement to add? I think the details approach you have already is the best one 😄

cjrace: Good spot - have updated that in the latest commit

sarahmwong (approved):
looks good to me thanks Cam!

PR #135, opened 2025-01-27
Fix cookies banner link in example app

cjrace:
## Overview of changes

Quick one that fixes #119, as I was bogged down in bigger chunks of work and wanted to pick off a nice easy one I knew I could do quickly!

As @sarahmwong suggested in the issue we were just missing the server `observeEvent()` function to watch for the link being pressed, added that so it now changes the tab to the cookies page.

## PR Checklist

- [ ] I have updated the documentation (if needed)
- [ ] I have added or updated tests for these changes (if applicable)
- [x] I have tested these changes locally using `devtools::check()`

## Reviewer instructions

Main thing is to run the test dashboard and check the link changes the page okay.

PR #223, opened 2026-05-12
Fix code scanning false-positives by running load_all and lint in one Rscript session

cjrace:
## Overview of changes

The repo currently has 5 open Security → Code scanning alerts, all from `lintr`'s `object_usage_linter` and all about the `attachDependency` internal helper:

| Alert | Location |
|---|---|
| #553 | `R/gov_list.R:74` |
| #557 | `vignettes/layout-options.Rmd:15` |
| #564 | `vignettes/shinyGovstyle.Rmd:15` |
| #567 | `vignettes/layout-options.Rmd:353` |
| #568 | `vignettes/shinyGovstyle.Rmd:83` |

`attachDependency` is a valid internal helper (`@noRd`) called ~30 times across the package — these are all **false positives**. The alert message itself says: _"Could not find exported symbols for package 'shinyGovstyle' … there is no package called 'shinyGovstyle'. This may lead to false positives."_

**Root cause.** The reusable workflow in `dfeshiny` runs `devtools::load_all()` and `lintr::lint_package()` in **two separate `Rscript {0}` steps**. Each step is a fresh R process, so the package namespace loaded in step 1 is gone by step 2. Without the package loaded, `object_usage_linter` can't see internal symbols.

**Fix.** Inline the workflow steps here and merge both calls into a **single `Rscript` invocation** so the loaded namespace is still in scope when lintr runs. A `tryCatch` around `load_all()` preserves the existing `continue-on-error: true` intent — SARIF is still uploaded if `load_all` fails, and the failure is visible in the logs.

Once confirmed green here, a matching fix will be raised upstream in [dfe-analytical-services/dfeshiny](https://github.com/dfe-analytical-services/dfeshiny) so the reusable workflow benefits all consumers. This repo can then revert to `uses: …lintr_reusable.yaml@main`.

## PR Checklist

- [x] I have updated the documentation (if needed) — no doc changes required
- [x] I have added or updated tests for these changes (if applicable) — CI workflow only, no unit tests
- [ ] I have tested these changes locally using `devtools::check()` — the SARIF upload path is CI-only; verify by watching the Actions run on this PR

## Reviewer instructions

1. Let the `lintr` Actions workflow run on this PR.
2. Check the **"Load package and lint in one session"** step: it should not print `load_all failed:` and should produce `lintr-results.sarif`.
3. After the run, check Security → Code scanning — alerts #553, 557, 564, 567, 568 should move to `fixed` state.


cjrace:
In theory won't be needed once https://github.com/dfe-analytical-services/dfeshiny/pull/135 is in

cjrace:
https://github.com/dfe-analytical-services/dfeshiny/pull/135 has fixed this, closing.

PR #224, opened 2026-05-12
Move to centralised reusable Air workflow

### LINKED ISSUE #160: Move to centralised, reusable Air workflow
Move to using a default Air workflow once https://github.com/dfe-analytical-services/dfeshiny/pull/116 is merged.

I have made an Air workflow directly into this repo for speed so we can reformat and have the check in place before doing more development next week.

We could leave this as is, but for maintenance and updates it's easier if we take advantage of a centralised script for it given we'll be using that for other packages and projects in DfE once launched.

## Additional context
Air docs are at - https://posit-dev.github.io/air/

cjrace:
## Summary

- Deletes the local `format-suggest.yaml` Air workflow
- Adds `air.yaml` as a thin caller that delegates to the shared [`dfe-analytical-services/dfeshiny/.github/workflows/air_reusable.yaml@main`](https://github.com/dfe-analytical-services/dfeshiny/blob/main/.github/workflows/air_reusable.yaml)

Closes #160

In addition, updates the CONTRIBUTING.md and adds an AGENTS.md with further documentation about contributing to the package.

## Behaviour changes

- Trigger changes from `pull_request` to `pull_request_target` (required for reviewdog to have write permissions to post suggestions)
- Workflow now also runs on push to `main`/`master` (harmless re-check after merge)


cjrace:
Have removed the old air format action from the required list in the repo settings and will add the new one after merging this one in (will do so once CI passes).

Issue #187, opened 2025-10-13
[Bug]: govTabs missing tabset ARIA role


cjrace:
### Overview

The govTabs function doesn't assign an ARIA role of tabset, so users relying on that aren't warned of the expected behaviour / highlighted to users when they saerch for tabs.

### Steps to Reproduce

Inspect the HTML of the showcase app.

### Expected Behavior

We need to assign `tablist` and `tab` roles within the govTabs function.

### Additional Information

User comments

“Although I found that the tabs element of the page worked as expected with JAWS, I noticed that the items which behaved like tabs, are identified as links by JAWS. Although this is not a significant problem for me as an advanced user of screen reading software, methods used to identify tabs can be different to those used to identify links. Including a clear indication to content and providing a consistent structure of navigation, will resolve this issue for users of screen reading software."

Page 23 of Sept 2025 accessibility audit report

PR #225, opened 2026-05-12
Wrap radio buttons in fieldset/legend for screen reader accessibility

### LINKED ISSUE #184: [Bug]: Accessibility - radio button fieldsets
### Overview

Using the component ‘Radio_button_input (inline)’ as an example on the showcase app, the ‘Yes’ and ‘No’ radio buttons are visually grouped under the label ‘Have you changed your name?’. When reading and interacting with this component using screen-reading software, this grouping is not relayed because the radio buttons are not contained within a fieldset with a legend to describe the group; this means that the relationships between the radio buttons may be difficult for a screen reader user to understand.

### Steps to Reproduce

Go to the radio button example using JAWs.

This also affects the download radios example.

### Expected Behavior

Comment from voice activation user

“Navigating the page with JAWS is confusing, as the series of radio buttons do not indicate any related information when using the ‘JAWS forms list’, or single letter navigation by pressing the ‘a’ key with JAWS activated. I found this confusing, as it is necessary to use the ‘arrow keys’ to identify the additional content. Including a consistent level of navigation by allowing users to identify the function of each option and any related content, will resolve this issue while navigating.”

We should ensure that visually grouped form elements are also grouped with a fieldset.

This ensures that screen reader users are aware of the relationship between form elements. The fieldset and legend elements can be used to group the content for screen reader users.

We have a couple of options, either a separate fieldset wrapper function we encourage developers to use around individual components (and add ourselves into the download_radios) or have something in the functions themselves to add this if necessary.

```
<fieldset>
 <legend>What's your favourite colour?</legend>
 <div>
 <label for="1blue">Blue</label>
 <input type="radio" name="colour" id="1blue" value="blue">
 <label for="1green">Green</label>
 <input type="radio" name="colour" id="1green" value="green">
 <div>
</fieldset>
```

### Additional Information

Page 15 of the September 2025 accessibility audit report.

cjrace:
Resolves #184 

## Overview

This started as a focused fix for #184 (radio buttons need a `<fieldset>` / `<legend>` so screen readers announce group context) but grew in scope once it became clear the same underlying pattern was also present in sibling components. 

The diff is large, but only three files contain logic changes (`R/utils.R`, `R/radio_button_input.R`, `R/checkbox_input.R`, `R/date_Input.R`); everything else is either generated docs or one-line `@inheritParams` swaps. Breakdown below.

## Details

1. The core fix (`R/radio_button_input.R`, `R/utils.R`)

`radio_button_Input()` now wraps its options in `<fieldset class="govuk-fieldset">` with the label rendered as `<legend class="govuk-fieldset__legend">`, so JAWS/NVDA/VoiceOver announce group context.
Extracted this into a shared internal helper, `govFieldset()`, since the next two items needed the identical structure.

Removed the now-unused `controlLabel2()` helper.

2. Why checkbox and date_Input are also touched

While building `govFieldset()` I found `checkbox_Input()` already had a `<fieldset>`, but it used a `<label>` inside it instead of a `<legend>` - the same accessibility bug. Rather than leave that half-fixed next to the new helper, I moved `checkbox_Input()` and `date_Input()` onto `govFieldset()` too, so all three grouped-input components now produce correct fieldset / legend markup and share one implementation instead of three near-duplicates.

This is also why `download_radios()` needed no changes: it's a thin wrapper over `radio_button_Input()` and inherits the fix automatically. It also covers the "download radios" case called out in the original issue.

3. New shared options

Moving three components onto one helper meant giving them the same params: label_size (legend size modifier) and heading_level (optional `<hX>` wrapping the legend, for the GDS "question as page heading" pattern) are now available on `radio_button_Input()`, `checkbox_Input()`, and `date_Input()`.

4. Standardising params

Adding label_size / heading_level / fieldset docs to three functions meant either repeating the same roxygen blocks everywhere or centralising them. I chose to centralise: shared parameter docs now live in `R/params.R` and are pulled in via `@inheritParams`.

5. Backfilling NEWS.md

Added in two NEWS.md entries (#187 and #189) that were merged after the v0.2.0 CRAN submission but never recorded - caught this while updating NEWS.md for this change and folded it in rather than opening a separate PR.

## Test plan

- [x] `devtools::test(filter = "radio")` — 28 tests pass
- [x] Run the showcase app and inspect HTML to confirm `<fieldset>`/`<legend>` structure


cjrace:
Ended up in a bit of a review hole with Claude, but have broke out of it now, tested it through and happy with it, will get it merged in as it touches so many docs.

PR #226, opened 2026-05-12
Add role="alert" to error messages for screen reader accessibility

### LINKED ISSUE #192: [Bug]: Error on form input not announced to screen readers
### Overview

Image

The service page has two textarea components. The second instance of a textarea in the section is used to test error handling. Activating the button ‘Check for errors’ causes an error message to appear. The error message reads ‘Please do not leave blank’. Those that use screen-reading software will not hear the software announce the error, and this is because the error message is not added to an element that uses the role=alert attribute.

### Steps to Reproduce

Trigger the error on the showcase app, inspect the HTML.

### Expected Behavior

Screen reader comments:

> “Once submitting the form with errors using JAWS, I became confused as no indication of any error handling was identified once the page had loaded. I found by using various methods of navigation, that errors were identified, however they do not follow the conventions of the Gov UK Design System. Including a clear error handling process will resolve this issue, causing less confusion for users of screen reading software."

We need to add

### Additional Information

Can review against the GOV.UK error message component - https://design-system.service.gov.uk/components/error-message/.

Page 38 of the September 2025 accessibility report.

cjrace:
## Summary

Fixes #192. Error messages on form input components were not announced by screen readers (JAWS/NVDA) when toggled on via `error_on()`, because the `<p class="govuk-error-message">` element had no live region.

- Added `role="alert"` to the error `<p>` element in all 7 input components: `text_Input`, `text_area_Input`, `checkbox_Input`, `radio_button_Input`, `date_Input`, `file_Input`, and `input_field`
- Updated corresponding tests to assert `role="alert"` is present, and corrected the `shinyjs-hide` attribute index (shifted by the new `role` attribute)
- Added NEWS.md entry

## Test plan

- [x] `devtools::test()` passes (282 tests, 0 failures)
- [x] Run the showcase app, navigate to the Service page, leave the second textarea blank and click **Check for errors** — inspect the DOM to confirm `role="alert"` on the error `<p>`
- [x] Verify with a screen reader (NVDA/JAWS on Windows) that the error text is announced when `error_on()` is triggered


cjrace:
Tested this with NVDA locally, the live app from main did not announce, but this does successfully announce the error as an alert as I hoped. Will merge in when CI are all passing

PR #227, opened 2026-05-19
Fix to link and regular text size

### LINKED ISSUE #194: [Bug]: Link text in contents is too small for the colour contrast

sarahmwong:
Replaces old PR #209 - built using the old CSS

## Overview of changes

Changed relevant instances so font size should be 1.2rem and font weight bold for any links

## PR Checklist

- [] I have added or updated tests for these changes (if applicable)

## Reviewer instructions

- [ ]  Check that the text and links look ok in the example app
- [x]  Check that css changes have been reflected accurately in the md file

@cjrace please can you also sense check that the header size small looks ok? I've been staring at this so long but I think it might look smaller than the main font now, but don't want to change it if it's actually ok!

### DIFF
--- a/R/download_link.R
+++ b/R/download_link.R
@@ -144,7 +144,7 @@ download_link <- function(
   # Create the link object
   link <- htmltools::tags$a(
     id = outputId,
-    class = "shiny-download-link disabled",
+    class = "shiny-download-link govuk-link disabled",
     href = "",
     target = "_blank",
     download = NA,
--- a/css_changes.md
+++ b/css_changes.md
@@ -11,6 +11,7 @@ color: #0b0c0c;
 
 * Add `font-size: 16px;` to root at top of file underneath `--govuk-frontend-version: VERSION NUMBER;`
 
+* Replace all instances of colour #1a65a6 with #175892
 
 * Change url links
 
--- a/inst/www/css/govuk-frontend-6.1.0.min.css
+++ b/inst/www/css/govuk-frontend-6.1.0.min.css
@@ -19,7 +19,7 @@
     --govuk-border-colour: #cecece;
     --govuk-input-border-colour: #0b0c0c;
     --govuk-hover-colour: #cecece;
-    --govuk-link-colour: #1a65a6;
+    --govuk-link-colour: #175892;
     --govuk-link-visited-colour: #54319f;
     --govuk-link-hover-colour: #0f385c;
     --govuk-link-active-colour: #0b0c0c;
@@ -73,7 +73,7 @@
 }
 
 .govuk-link:link {
-    color: var(--govuk-link-colour, #1a65a6)
+    color: var(--govuk-link-colour, #175892)
 }
 
 .govuk-link:visited {
@@ -142,7 +142,7 @@
 
 .govuk-link--no-visited-state:link,
 .govuk-link--no-visited-state:visited {
-    color: var(--govuk-link-colour, #1a65a6)
+    color: var(--govuk-link-colour, #175892)
 }
 
 .govuk-link--no-visited-state:hover {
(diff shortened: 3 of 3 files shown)


cjrace (changes requested):
Hey @sarahmwong, thanks for this!

I've had a test and I think the headers are okay actually, though I'm looking at this compared to main and then also GOV.UK and I'm no longer sure increasing the text size is the right move (aware I suggested it in the initial issue - I'm really sorry!). I think it might be deviating too far from the GOV.UK design system defaults.

I've been going back and forth on this a bit, and I think I might now have an alternative we could do instead that's a smaller deviation and still fixes the issue. To help with the link text meeting AAA, we could darken the link text colour slightly to something like #175892, which would have greater than a 7:1 ratio and meet AAA standard for contrast without being a major visual change like the size, and is still lighter than the hover colour so there's a bit of a change still there for visual users. Part of my thinking is also if people using the package have carefully designed a dashboard page to perfectly fit the size of their desktop window and then they update the package version it could all go awry with bigger fonts.
- If doing this we'd also need to add the govuk-link class to the download_link.R function, as when testing it out I realised that link doesn't inherit the GOV.UK link colours properly.

I think the best next step is to (either on a new branch or reverting on this branch), update the root line to:
```
    --govuk-link-colour: #175892;
```
Add a note into css_changes.md, including a comment that we're deviating to get greater than 7:1 contrast for links to meet AAA.

and then in R/download_link.R, line 147, we update the classes to:
```
    class = "shiny-download-link govuk-link disabled",
```

What do you think to this? Looking at the original issue the main problems in the screenshots were link text (which this should fix) and the contents links (which we've now dropped)

sarahmwong:
Sounds like a much neater option! I didn't like the look of the huge font (even with my terrible eyesight) deviating from gov.uk so much so this feels like a better fix - I'll take a look at adding in. Is it worth us flagging with the gov.uk design team too as think this is a problem with the design system itself?

cjrace:
> Sounds like a much neater option! I didn't like the look of the huge font (even with my terrible eyesight) deviating from gov.uk so much so this feels like a better fix - I'll take a look at adding in. Is it worth us flagging with the gov.uk design team too as think this is a problem with the design system itself?

Good question, I think for this as it's a AAA criteria, it's probably not worth it, as I believe GOV.UK only aims for AA standard instead, so they will likely be aware and deem it out of scope - we're aiming for a higher standard than they are.

sarahmwong:
Makes sense thanks @cjrace! I have tried to rebase and make those changes here now - please can you check that looks and runs ok on your end?

cjrace (approved):
Nice, thanks @sarahmwong - all checks out nicely and good catch on the CSS, I didn't realise how many places that colour was referenced!

Issue #178, opened 2025-10-03
Add default text into the `banner()` function


cjrace:
## Describe the solution you'd like
A feedback_url argument, NULL by default for backwards compatibility, but if a URL is given then the text is automatically populated with:

>This is a new service - your [feedback (opens in new tab)](URL) will help us to improve it.

e.g.
<img width="518" height="41" alt="Image" src="https://github.com/user-attachments/assets/a3380823-c171-43c3-8283-25945b99b15c" />

## Describe alternatives you've considered
Considered setting this as the default behaviour for the function, but wary it's probably not worth the potentially breaking change.

Could also make it a bit fancier and have it automatically recognise if the URL is a mailto: link, and in that case provide appropriate text, e.g.

>This is a new service - please contact [email...](mailto:...) if you have any questions or feedback.

## Additional context
Currently in our showcase app, and many dashboards using shinyGovstyle we have similar text being used, though to do that you have to put the HTML in and paste it. Add this feature would only be a small thing but it would clean up code for some dashboards and generally set more of a standard, consistent text for people to use.

```
banner(
  "banner",
  "Beta",
  paste0(
    'This is a new service \u002D your <a class="govuk-link" href=',
    '"https://github.com/dfe-analytical-services/shinyGovstyle/issues/new',
    '/choose">feedback</a> will help us to improve it.'
  )
),
```

PR #230, opened 2026-05-26
Allow taglists for text content

### LINKED ISSUE #221: [Feature]: Allow html tag objects to be passed to inset_text() aka insert_text()
### Description

Currently `insert_text(...)` will only accept plain string input, but should take html tags.

As a basic example:
```
shinyGovstyle::insert_text(inputId = "bold-text", text = shiny::tags$b("Bold text"))
Error in `FUN()`:
! argument is not a character vector
```

What I'd like is for the `text` argument to accept a range of things along the lines of:
`shinyGovstyle::insert_text(inputId = "bold-text", shiny::tags$b("Bold text"))`
`insert_text(inputId = "bold-text", tagList(shiny::tags$b("A link: "), external_link("link-url", "link text")))`

Also, should the parameter be renamed given it's not just text?

### Use case

...

### Why this is important

...

### Additional Context

...

cjrace:
## Overview of changes

Resolves #221.

Allow text-content arguments across several components to accept `shiny` tag objects (e.g. `shiny::tags$b("Bold")`) and`shiny::tagList()` values, not just plain character strings. This makes it possible to compose more details content (e.g. with bold, links, line breaks, nested tags) without resorting to raw HTML strings.

  Components updated:
  - `insert_text()` — `content` (renamed from `text` as suggested in #221; old arg deprecated via `lifecycle`)
  - `panel_output()` — `sub_text`
  - `noti_banner()` — `body_txt`
  - `details()` — `help_text`
  - `banner()` — `label`
  - `warning_text()` — `text`
  - `gov_summary()` — `info`

A small internal helper `as_govuk_html()` (in `/utils.R`) centralises the logic: tag / tagList / HTML input passes through unchanged, plain strings continue to be wrapped with `shiny::HTML()` so existing behaviour is preserved and therefore this change should be backwards compatible.

## PR Checklist

## Reviewer instructions

Manually poke at a few of the changed functions and test the use case referenced in #221.

Test out the raw HTML still working, example in #178 for the beta banner.

Also tagging @rmbielby for sight, given it was their initial idea / request.

### DIFF
--- a/tests/testthat/test-warning_text.R
+++ b/tests/testthat/test-warning_text.R
@@ -17,12 +17,12 @@ test_that("text accepts a tagList", {
     "warningId",
     shiny::tagList(
       "You can be fined up to ",
-      shiny::tags$strong("£5,000"),
+      shiny::tags$i("£5,000"),
       " if you do not register."
     )
   )
 
   html <- as.character(out)
   expect_match(html, "You can be fined up to", fixed = TRUE)
-  expect_match(html, "<strong>£5,000</strong>", fixed = TRUE)
+  expect_match(html, "<i>£5,000</i>", fixed = TRUE)
 })
--- a/tests/testthat/test-warning_text.R
+++ b/tests/testthat/test-warning_text.R
@@ -1,8 +1,28 @@
-test_that("warning works", {
-  warn_check <- warning_text("warningId", "Test")
+test_that("warning renders string text in the body", {
+  out <- warning_text("warningId", "Test")
 
-  expect_identical(
-    warn_check$children[[2]]$children[[1]],
-    "Test"
+  html <- as.character(out)
+  expect_match(html, "govuk-warning-text", fixed = TRUE)
+  expect_match(html, "Test", fixed = TRUE)
+})
+
+test_that("text accepts a shiny.tag", {
+  out <- warning_text("warningId", shiny::tags$b("Bold warning"))
+
+  expect_match(as.character(out), "<b>Bold warning</b>", fixed = TRUE)
+})
+
+test_that("text accepts a tagList", {
+  out <- warning_text(
+    "warningId",
+    shiny::tagList(
+      "You can be fined up to ",
+      shiny::tags$strong("£5,000"),
+      " if you do not register."
+    )
   )
+
+  html <- as.character(out)
+  expect_match(html, "You can be fined up to", fixed = TRUE)
+  expect_match(html, "<strong>£5,000</strong>", fixed = TRUE)
 })
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ docs/
 *_.new.png
 .claude/
 inst/doc
+AGENTS.md
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: shinyGovstyle
 Title: Custom Gov Style Inputs for Shiny
-Version: 0.2.0
+Version: 0.2.0.9000
 Authors@R: c(
     person("Ross", "Wyatt", , "ross.wyatt@justice.gov.uk", role = "aut"),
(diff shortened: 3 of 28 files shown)


tests/testthat/test-warning_text.R
```diff
@@ -1,8 +1,28 @@
...
+
+  expect_match(as.character(out), "<b>Bold warning</b>", fixed = TRUE)
+})
+
+test_that("text accepts a tagList", {
+  out <- warning_text(
+    "warningId",
+    shiny::tagList(
+      "You can be fined up to ",
+      shiny::tags$strong("£5,000"),
```
sarahmwong: change test to tags$i? Think warning text outputs all in bold/strong regardless of the shiny tag - doesn't really matter as we're just testing functionality I guess, but handy to see it in action!

tests/testthat/test-warning_text.R
```diff
@@ -1,8 +1,28 @@
...
+      " if you do not register."
+    )
   )
+
+  html <- as.character(out)
+  expect_match(html, "You can be fined up to", fixed = TRUE)
+  expect_match(html, "<strong>£5,000</strong>", fixed = TRUE)
```
sarahmwong: and change expected output

sarahmwong (changes requested):
Thanks Cam this looks great! Can confirm this solves the use case in #221 and that existing html inputs still work. Teensy change to one of the tests as I think the warning outputs as strong regardless of what we put in there (though know the test is working as it should do just testing for a strong within a strong!) - just thought it's handy to be able to see ::smile::

tests/testthat/test-warning_text.R
cjrace: Ahhh, great spot!

cjrace: Fixed this now so it's clearer

tests/testthat/test-warning_text.R
cjrace: Fixed

cjrace:
Thanks @sarahmwong, made those tweaks and also bumped us to 0.2.0.9000 as a dev version so the main branch is distinct from the version on CRAN now we're starting to line up changes for the next release

sarahmwong (approved):

rmbielby:
One thing @cjrace @sarahmwong, did you look at doing this for `gov_list()` as well at all?

cjrace:
> One thing @cjrace @sarahmwong, did you look at doing this for `gov_list()` as well at all?

No, hadn't considered it - but good thinking! I'm assuming this is for adding in links to the bullet points?

rmbielby:
> > One thing @cjrace @sarahmwong, did you look at doing this for `gov_list()` as well at all?
> 
> No, hadn't considered it - but good thinking! I'm assuming this is for adding in links to the bullet points?



> > One thing @cjrace @sarahmwong, did you look at doing this for `gov_list()` as well at all?
> 
> No, hadn't considered it - but good thinking! I'm assuming this is for adding in links to the bullet points?

Yep, exactly. Sorry, was one I had in mind early on, but forgot to name it explicitly in the original issue.

cjrace:
> > > One thing @cjrace @sarahmwong, did you look at doing this for `gov_list()` as well at all?
> > 
> > No, hadn't considered it - but good thinking! I'm assuming this is for adding in links to the bullet points?
> 
> 
> 
> > > One thing @cjrace @sarahmwong, did you look at doing this for `gov_list()` as well at all?
> > 
> > No, hadn't considered it - but good thinking! I'm assuming this is for adding in links to the bullet points?
> 
> Yep, exactly. Sorry, was one I had in mind early on, but forgot to name it explicitly in the original issue.

No need to apologise, I should have picked it up when reviewing where to apply it. I have a draft branch locally I'll get up for review in the next week or so 👍

PR #231, opened 2026-05-27
Add update page title function

### LINKED ISSUE #185: [Bug]: Make page title update on showcase app
### Overview

Currently we have a static page title, however we have multiple 'pages' within that app that have separate H1 headers, which then go out of sync with the page title. The lack of change in page title means that not all users are notified of the changes of content on the page.

### Steps to Reproduce

Go to the showcase app, and inspect the title element in the HTML.

### Expected Behavior

We should update the page title each time the h1 changes, e.g.

shinyGovstyle component showcase | tables and tabs
shinyGovstyle component showcase | text inputs

Maybe also worth thinking about condensing the initial title so it's more concise too - though need to make sure that's not at the cost of understandability.

### Additional Information

Can either pull in another dependency like https://github.com/ashbaldry/shinytitle/, or do our own title update function using JS (the latter might be better for long term maintainability, and it's pretty simple to add), e.g.

```
document.title="New Title";
document.getElementById("readTitle").innerHTML = document.title;
```

cjrace:
## Overview of changes

Keeps the browser tab title in sync with the visible page in multi-page Shiny apps, a GOV.UK Design System accessibility recommendation, since screen readers announce the page title on navigation. Resolves #185.

### Function updates
- `update_page_title()`: new function to set the browser tab title from the server, with optional service_name to produce the GOV.UK-recommended "Page | Service" format.
- `service_navigation()`: now syncs the tab title automatically when a user clicks a nav link (new `auto_page_title` and `page_title_suffix` arguments, on by default).
- `service_navigation_server()`: wires every nav link to its tab panel in one call, replacing per-link `observeEvent()` boilerplate.
- `navigate_to()`: one-call helper for programmatic navigation (next / back buttons, footer shortcuts, modal links) that switches the tab panel and updates the active nav link together.

### Behaviour change

`service_navigation()` now updates the browser tab title by default. Apps that need the old behaviour can opt out with `auto_page_title = FALSE`. Noted in NEWS.md.

- Example app updated to demonstrate the new flow.
- Tests added for `update_page_title()`, `navigate_to()`, and the new `service_navigation()` behaviour.
 
## PR Checklist

## Reviewer instructions

Check that the functions are easy to use and the documentation is easy to follow.

Give the example app a test and also check the updated code - is this easy for analysts to implement themselves?

Review the test cases to see if the coverage covers the things you'd expect to see for this.


sarahmwong (approved):
Thanks Cam happy to approve, works really neatly and use and documentation looks clear to me!

Issue #229, opened 2026-05-23
Refactor tests to remove positional $children[[2]]$children[[1]]...


cjrace:
### Description

Currently the tests tend to rely on very specific positional calls such as `$children[[2]]$children[[1]]...`. I'd like to remove them and instead check for specific elements / roles, making the structure more flexible mirroring how a real user / assistive tech would navigate by class / id.

### Why this is important?

Currently the tests are very brittle and can break if we want to change anything about the HTML structure, such as adding or removing elements, even if behaviourally this doesn't change anything. 

Changing to test by class / id / attribute instead of position would also make the tests more readable and make it easier to understand what expected behaviour we're testing for, instead of being a detailed test of the specific HTML structure.

PR #233, opened 2026-05-28
191 bug accordions dont fully announce state

### LINKED ISSUE #191: [Bug]: Accordions don't fully announce state
### Overview

Users will encounter accordions beginning at ‘Show all sections’ and ending at ‘How people read’. Encountering these accordions with screen-reading software, a screen reader user will not hear a collapsed or expanded state. The software is unable to announce a state because the aria-expanded attribute is omitted. This means that users will not know the current state.

### Steps to Reproduce

Inspect the HTML on the showcase app.

### Expected Behavior

`aria-expanded="false"` and `aria-expanded="true"` should be added to each accordion within the button HTML, e.g.

```
<button aria-expanded="false" class="govuk-accordion__section-button" 
id="accordion-default-heading-01" name="accordion-default-heading-01">
 <span class="govuk-accordion__section-heading-text">
 <span class="govuk-accordion__section-heading-text-focus">Writing well for the web</span>
 </span>
 <span class="govuk-visually-hidden govuk-accordion__section-heading-divider"></span>
 <span class="govuk-accordion__section-toggle">
 <span class="govuk-accordion__section-toggle-focus">
 <span class="govuk-accordion-nav__chevron govuk-accordion-nav__chevron--
down"></span>
 <span class="govuk-accordion__section-toggle-text">
 Show
 <span class="govuk-visually-hidden">this section</span>
 </span>
 </span>
 </span>
</button>

```

### Additional Information

Page 34 of the September 2025 accessibility report

### LINKED ISSUE #228: [Bug]: Accordion behaviour replicates on all accordions
### Overview

Changes to accordions replicate on all accordions across all pages. For example, a 'hide all' or 'show all' will hide or show all. Hiding or showing specific accordion will replicate on all different accordions for the same number section (for example showing the first section on an accordion will show the first section on all accordions).

This is because thaccordion.js is currently written to pick up all instances of an accordion number or class in a document. 

### Steps to Reproduce

Edit example app to have multiple accordions to see how each accordion changes when the other is set to hide or show.

### Expected Behavior

All accordions should operate independently of each other, so when one accordion is shown or hidden it does not impact the other accordions. 

### Additional Information

Fix for #191 will need to be fixed in this way.

jenisayshi:
## Overview of changes

Changes to accordions are:

1. Get accordions to work independently of each other (e.g a change to on accordion on a page will not affect another). Fixes #228 
2. Adds aria-expanded attribute to both show all/hide all controls and each section. This updates to true/false on click of both sections and show all/hide all. Fixes #191 
3. Rewrite code and tests to use class and element sectors instead of position selectors. Linked to #229 

## PR Checklist

## Reviewer instructions

1. Manually test multiple accordions between tabs (and on the same tab) to check accordions work independently of each other. 

2. Test that aria-expanded shows correct value:
(i) by section when individual section changes
(ii) in controls when hide/all show all changes
(iii) for every section when hide all/show all changes
(iv) that (i)-(iii) work independently between accordions

### DIFF
--- a/R/accordion.R
+++ b/R/accordion.R
@@ -60,7 +60,6 @@ accordion <- function(
           class = "govuk-accordion__show-all",
           `aria-expanded` = "false",
           shiny::tags$span(
-            id = "show-all-chevron",
             class = paste(
               "govuk-accordion-nav__chevron",
               "govuk-accordion-nav__chevron--down"
@@ -90,6 +89,7 @@ accordion <- function(
                   class = "govuk-accordion__section-button",
                   id = paste0("accordion-default-heading-", z_str),
                   name = paste0("accordion-default-heading-", z_str),
+                  `aria-expanded` = "false",
                   shiny::tags$span(
                     class = "govuk-accordion__section-heading-text",
                     shiny::tags$span(
--- a/inst/www/js/accordion.js
+++ b/inst/www/js/accordion.js
@@ -1,104 +1,109 @@
-// individual section buttons
+// On click of individual sections
 $(document).on('click', '.govuk-accordion__section', function (e) {
 
-  // get 'name' value from button
-  var str = e.target.name;
-
-  // get z value from name (Map argument)
-  var level = str.substr(str.length - 2);
-
-
-
   // get class from top div
-  var cur_class = $('[name="accordion-default-heading-' + level + '"]').parent().parent().parent()[0].classList.value
+  var top_div = e.target.closest('.govuk-accordion__section');
+
+  var cur_class = top_div.classList.value;
+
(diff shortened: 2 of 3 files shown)


sarahmwong (changes requested):
Thanks for this Jeni! Have tested the first review instruction but the "show all" no longer works on the example app - think this functionality has been lost in this fix but please could we add back in?

cjrace:
Hey @jenisayshi / @sarahmwong - I think I've fixed the issue you had Sarah, and I've gone through and sorted out the merge conflicts with the other updates I've been making so I think this one is good to go now!

PR #234, opened 2026-05-30
Refactor tests to navigate by classes instead of position

### LINKED ISSUE #229: Refactor tests to remove positional $children[[2]]$children[[1]]...

cjrace:
## Overview of changes

Reworks the test suite to remove brittle patterns and more deliberately test the specifics we care about. Previously the tests were overly strict on structure, meaning they could fail unnecessarily from dependency changes or just us adding elements into components, while also being hard to read and understand what exactly we're trying to test.

Also documented the approach for ease moving forwards. Used Claude to assist with this one as it would have taken too long to work it out myself to be worth it otherwise!

On top of this, I've also taken the chance to expand the test coverage on existing functions.

### What changed
- Adds helper-tags.R — shared helpers for navigating tag trees by class / id (`find_tag()`, `find_tags()`, `find_tag_required()`, `tag_text()`, `child_classes()`, `find_by_id_suffix()`, `expect_hidden_error()`, `expect_has_tag()`, `expect_no_tag()`) instead of chaining `$children[[N]]`.
- Adds helper-shinyjs.R — a mock recorder for testing `shinyjs::*` side effects without a Shiny session.
- Replaces several snapshots (error_summary, external_link, file_Input) with explicit code-based assertions. Keeps snapshots only for components that embed external renderers (`actionLink()` / {reactable}).
- Adds error/warning path coverage (test-error_on, test-error_off, test-error_summary_update, test-govLayouts).
- Documents the testing philosophy and snapshot policy in CONTRIBUTING.md.

## PR Checklist

## Reviewer instructions

Tests only, no changes to package behaviour, so a pretty safe change. Main thing is to review the testing approach details in the documentation to make sure it makes sense (both are the docs easy to follow, and do you agree or have any comments on the approach taken!).

Tagging both @jenisayshi and @sarahmwong for awareness / interest as I think you both are working on branches actively and have seen the issue this resolves #229.


cjrace:
Finally got this one ready for you to check over @sarahmwong! For reviewing, as it's quite a lot of repetitive lines, I'd focus on checking that all the package behaviour is as you'd expect still in the example app, and then judging how readable the tests and the notes about the testing philosophy are

sarahmwong (approved):
Ahhh wonderful thank you for looking at this beast Cam! Happy that the contributing file gives clear instructions, new helper functions for testing are really useful and tests are much tidier now. Functionality of the error messages in example app still works as expected.

PR #133, opened 2025-01-23
make accordian open/close by clicking on any part of the section

jenisayshi:
## Overview of changes

Update accordion to open when you click anywhere on the accordion in line with govuk. Fixes #123.

## PR Checklist

- [ ] I have updated the documentation (if needed)
- [ ] I have added or updated tests for these changes (if applicable)
- [x] I have tested these changes locally using `devtools::check()`

PR #235, opened 2026-06-01
Add update radio server helper and cookies vignette

### LINKED ISSUE #67: Check if there's any updates to the cookies banner from dfeshiny
In [dfeshiny](https://github.com/dfe-analytical-services/dfeshiny) we're making a specific module for the cookies banner that links to Google Analytics. It's worth checking we we've added there to see if there's any updates to make to the cookies function available in this package.

cjrace:
## Overview of changes

Adds **`update_radio_button_Input()`**, which allows you to update a Govstyle radio's selected option, choices, or label from the server, just like `shiny::updateRadioButtons()`. Works inside modules (auto-namespaced).

Also fixes #67, and adds a **"Cookies and analytics" vignette** showing how to keep a cookies-settings radio in sync with the cookie banner.

Also adds an example of this into the showcase app (a cookies page with a radio option that updates the same value that the banner updates).

### Why

There was previously no way to change a Govstyle radio from the server as @rmbielby highlighted in https://github.com/dfe-analytical-services/dfeshiny/pull/133#discussion_r3325584482. This meant you couldn't, for example, reflect a cookie banner choice on a settings page. This adds that and makes the client side actually apply the change.

## PR Checklist

## Reviewer instructions

- Test out the server side updating of radio buttons
- Test out the example put into the example app
- Review the docs to make sure they make sense and cover everything you'd expect


sarahmwong (approved):
Very nice! - thank you Cam looks good to me.

PR #236, opened 2026-06-04
Extend taglist content to more components

cjrace:
## Overview of changes

As highlighted by @rmbielby in #230, there were still more components that only accepted plain text where you'd reasonably want links or bold markup. Now you can now pass HTML strings, shiny tags (e.g. shiny::tags$b("Bold text")), and shiny::tagList() more widely across all components as advised in the GOV.UK design system.
  
- gov_list() and accordion() accept tags/tag lists for list items and section content. Accordion sections can now hold block content (lists, multiple paragraphs), not just a single paragraph.
- Labels and hints on label_hint(), text_Input(), text_area_Input(), date_Input(), select_Input(), checkbox_Input(), and radio_button_Input() now consistently accept plain strings, HTML strings, tags, and tag lists.

Backwards compatible; existing plain-string usage is unchanged.

I found an issue in our reusable air GitHub action where it was checking the main branch code instead of the PR branch, so have switched that back with a comment.

## PR Checklist

## Reviewer instructions

Check the places we're allowing HTML / styled text content are acceptable, and that we're not missing any other places that we should be making this available.


sarahmwong (approved):
Looks good, reviewed and think this now covers all components we could assume a user would want to sensibly add a link. Like the param control I've not seen that before!

PR #238, opened 2026-06-06
Fix nested recursion and reverse ordering bugs in tables

### LINKED ISSUE #237: [Bug]: Confusing error message for large tables
### Overview

Hello, shinyGovstyle has been great and very helpful for us at DHSC. Just had a minor issue where I was testing a table and getting the error message:
Error: evaluation nested too deeply: infinite recursion / options(expressions=)?

After some investigation, this was fixed once I reduced the table size below 1200 rows. Is it possible to add a more helpful error message in this case?

### Steps to Reproduce

Add a shinyGovtable with more than 1200 rows

### Expected Behavior

More informative error message

### Additional Information

_No response_

cjrace:
## Overview of changes

Fixes #237.

### What was wrong

`govTable()` failed with a confusing error (evaluation nested too deeply: infinite recursion) once a table grew past ~1,200 rows, as reported in [#237](https://github.com/dfe-analytical-services/shinyGovstyle/issues/237).

Bonus: While fixing the original issue, I noticed the row order change and realised that the old version was returning rows back in reverse order, which feels like a bug we've just not spotted or realised yet.

### What this changes

Reworked the row building to be more efficient so that large tables now render, and the confusing error message for users no longer appears.

Also added a warning for bigger tables. Tables over 1,000 rows now give a warning suggesting `govReactable()` instead of `govTable()`, which is paginated, sortable, and searchable, and better optimised for larger tables.

Bonus: Rows now appear in the right order. They were previously rendered in reverse (a long-standing bug, hidden by small example tables).

Tests: Added additional tests covering the warning, row order and larger tables to prevent this happening again.

## PR Checklist

## Reviewer instructions

For @sarahmwong in particular, two decisions I've made that I want to check your thoughts on:

1. The 1,000 row point as the trigger for the warning, it's relatively arbritrary currently!

2. The row-order fix is a visible behaviour change for anyone who happened to rely on the old reversed output (none should). It's a breaking change but one I think is better / would be the expected behaviour so I've made it and I've referenced it as a breaking change in NEWS.md. I tried to look for reasons why it might have been built that way initially, but couldn't come up with any reason not to fix the ordering to match the input data.frame. Let me know if you disagree!


sarahmwong:
Thanks for fixing Cam! Agree it makes sense as a breaking change. I wonder if we should reduce that warning list down even further, maybe to 100 or even 50 rows? 1000 rows on a page still seems quite long and I can't think of a use case where that would be preferable over a reactable table/little preview and option to download the data in a different file type

cjrace:
> Thanks for fixing Cam! Agree it makes sense as a breaking change. I wonder if we should reduce that warning list down even further, maybe to 100 or even 50 rows? 1000 rows on a page still seems quite long and I can't think of a use case where that would be preferable over a reactable table/little preview and option to download the data in a different file type

Thanks @sarahmwong - agree! Think I was overly cautious initially, have dropped the warning trigger to be 50 😄

sarahmwong (approved):
SORRY THIS HAS BEEN SO SLOW! I now have access to github at work though so can be less annoying going forward (sorry Cam!!!)

PR #244, opened 2026-08-11
Fix/193 character count is not announced to screen readers

### LINKED ISSUE #193: [Bug]: Character count is not announced to screen readers
### Overview

Character count isn't announced to screen reader users.

### Steps to Reproduce

Inspect the HTML of the showcase app

```
<div class="govuk-hint govuk-character-count__message">
 <span>You have used</span>
 <span id="text_area2wc">0</span>
 <span id="text_area2wl">of the 300 allowed</span>
</div>
```

### Expected Behavior

Screen reader comments:

> “While navigating the page with NVDA and entering content into the text area input field just before the ‘check for errors button,’ I found that on using the ‘arrow keys’ to explore the page identifies a character count which is not indicated for users of screen reader software while typing. I found this confusing, as it is easy to miss this information as it appears after the input field, and if tabbing to the next control is not accessed. Including a clear indication to the character count by announcing this after the first few words typed, or if a pause in typing has occurred for example, will resolve this issue for users of screen reading software.”

### Additional Information

Can review against the GOV.UK character count component - https://design-system.service.gov.uk/components/character-count/.

Page 40 of the September 2025 accessibility audit report.

sarahmwong:
## Overview of changes

Switched from using shiny counter to js instead to count the words - this should work like how the gov uk design system text counter works now and my screenreader is announcing the word count and countdown. Tweaked the wording slightly so it matches what we have in the design system to simplify the language. Also now errors and prints a ' x words over' error message when they go over the number of words allowed. 

Kept tests as they are but know there's wider work in #229 to improve these (and very much agree they should change!)

## PR Checklist

## Reviewer instructions

Check that the functionality works as expected for you

If you have access to any screenreader tools outside of Edge and the chrome extension please check to see if the word count is announced there too (have checked with both of these and seems fine!)


inst/www/js/text_area_input.js
```diff
@@ -0,0 +1,45 @@
...
+    var maxWords = parseInt($container.attr('data-maxwords'), 10);
+    if (isNaN(maxWords)) {
+      return;
+    }
+
+    var wordCount = countWords($textarea.val());
+    var $info = $('#' + $textarea.attr('id') + '-info');
+
+    if ($info.length) {
+      $info.text('You have ' + (maxWords - wordCount) + ' words left');
```
cjrace: Should this be 'remaining' instead of 'left' to match - https://design-system.service.gov.uk/components/character-count/

R/text_area_input.R
```diff
@@ -39,11 +39,37 @@ text_area_Input <- # nolint
...
+      },
       id = paste0(inputId, "div"),
-      shiny::tags$label(as_govuk_html(label), class = "govuk-label"),
-      shiny::tags$div(as_govuk_html(hint_label), class = "govuk-hint"),
+      `data-module` = if (!is.null(word_limit)) {
+        "govuk-character-count"
+      },
+      `data-maxwords` = if (!is.null(word_limit)) word_limit,
+      shiny::tags$label(
+        shiny::HTML(label),
```
cjrace: Was it intentional to drop the `as_govuk_html()`? I think it should be here


```suggestion
        as_govuk_html(label),
```

R/text_area_input.R
```diff
@@ -39,11 +39,37 @@ text_area_Input <- # nolint
...
+        class = "govuk-label",
+        `for` = inputId
+      ),
+      if (!is.null(hint_label)) {
+        shiny::tags$div(
+          hint_label,
```
cjrace: Should we wrap hint_label in `as_govuk_html()` (it used to be wrapped in it)?

cjrace (changes requested):
Hey @sarahmwong! Thank you for this, think it'll be a nice improvement to do it this way 😀 tested it on NVDA and it announced as I expected initially - though I think it's not fully like GOV.UK yet

Couple of comments in line, plus:

1) I think `word_count()` is now no longer needed, should we deprecate it and then also remove the reference from `vignettes/shinyGovstyle.Rmd`?

2) Can we now drop lines 80-84 of the `/example_app/modules/mod_text_types.R` script? Not sure we need that observeEvent anymore?

3) Could you add a note into the NEWS.md for this fix?

4) The initial announcement gives the title for the item then the hint then the word limit, but on GOV.UK's component it seems to go title then the word limit then the hint?

5) On GOV.UK's version in NVDA it announces the remaining characters / words when you stop typing, but this version does do that, I think it's something to do with needing to mark it as an aria-live region, or something in the JS - happy to look into it more if helpful?

6) Minor thing, we could give the singlular 'word remaining' for when there's only 1 word left, if you fancied a bit more JS work (not the most important if not though!)

7) Agree on the tests, happy with what you have here and can do the fuller change on the other branch

8) The error formatting doesn't quite match the GOV.UK one (left on this screenshot versus this branch on the right) - I'm easy on this, as what you've added is already a clear improvement and this PR was mainly for the screen reader side, just thought I'd mention as I noticed it while testing

Image

cjrace:
Think I've managed to fix this up now @sarahmwong! 

I've fixed up the styling (credit to Claude!), updated to mirror the GOVUK Frontend approach for announcing the word count status with the aria-live, while debouncing the announcement so it only announces when you stop typing.

(Extra commit below as well tidying up a couple of things I missed by my custom Claude reviewer pulled me up on!)

sarahmwong:
> Think I've managed to fix this up now @sarahmwong!
> 
> I've fixed up the styling (credit to Claude!), updated to mirror the GOVUK Frontend approach for announcing the word count status with the aria-live, while debouncing the announcement so it only announces when you stop typing.
> 
> (Extra commit below as well tidying up a couple of things I missed by my custom Claude reviewer pulled me up on!)

Thanks for this Cam!! This looks good - tested and happy with the functionality and edits to the tests you've made in line with the wider testing changes. If you can accept the requested changes am happy to merge/happy if you want to do that!

cjrace:
Awesome - thanks @sarahmwong, I'll approve / merge this in 😄

cjrace (approved):

PR #248, opened 2026-09-11
243: govReactable custom column formatting

### LINKED ISSUE #243: [Feature]: Support column-specific formatting in govReactable()
### Description

I've switched from reactable() to govReactable() for a DfE dashboard to align with GDS styling. In reactable(), I have previously used the columns argument to apply formatting to individual columns. For example, displaying one numeric column to 1 decimal place while other numeric columns are displayed to 0 decimal places. I haven't been able to find an equivalent approach in govReactable(). When I attempted to pass a columns argument, I received:
 _Error: formal argument "columns" matched by multiple actual arguments_ 
Am I missing an existing way to apply column-specific formatting in govReactable(), or is this functionality currently unavailable?
If it isn't currently supported, would support for column-specific formatting be a useful enhancement?

### Use case

One use case is displaying counts and percentages within the same table.

Percentage values often need to be displayed to a fixed number of decimal places for consistency. For example, values such as 4.0% and 4.7% should both display to one decimal place. Without column-specific formatting, trailing zeros are omitted, resulting in inconsistent presentation (for example, 4% alongside 4.7%).

Currently, formatting can only be applied consistently across all columns. Applying a single decimal format to the entire table results in counts being displayed as 1,234.0 and 56.0, while applying no decimal formatting causes percentage values to lose their consistency.

### Why this is important

If column-specific formatting is already supported, some additional documentation or examples would be helpful. If it is not currently supported, adding this functionality could make govReactable() more flexible for dashboards that contain a mix of measures requiring different display formats.

### Additional Context

Happy to provide a reproducible example if helpful.

cjrace:
## Overview of changes

Resolves #243.

`govReactable()` now takes a columns argument, so you can customise individual columns. For example, fixing decimal places in only one column:
```
govReactable(
  head(airquality[, c("Temp", "Wind")]),
  columns = list(
    Wind = reactable::colDef(format = reactable::colFormat(digits = 1))
  )
)
```

## PR Checklist

## Reviewer instructions

Will wait for the original issue raiser to confirm it works / give us a reproducible example to test before merging.


cjrace:
Tested this with @amsparks and seems to fix it! Merging in now so it can go in the next release

PR #249, opened 2026-09-12
Add an example of a list of links that open in a new tab

### LINKED ISSUE #195: [Bug]: Give a better example for add_warning = FALSE link text
### Overview

In the example showcase app we include the `add_warning = FALSE` example within a text paragraph, which is not the ideal placement.

Instead we should create a list of links, that has a note at the top warning that the links will open in new tabs.

We could / should also add an option to include an icon for visual users, as shown in https://www.w3.org/WAI/WCAG22/working-examples/css-new-window-warn/.

### Steps to Reproduce

Image

On the showcase app

### Expected Behavior

A warning for visual users that the link will open in a new site (the screen reader warning was there and worked well).

### Additional Information

Further advice and examples can be found at https://www.w3.org/WAI/WCAG22/Techniques/general/G201 and https://www.w3.org/WAI/WCAG22/working-examples/css-new-window-warn/ (icon warning we could investigate using as a third option).

Page 47 of the September 2025 accessibility audit report.

cjrace:
## Overview of changes

Resolves #195 (note that another PR will pick up the optional icon that can be used for new tab).

Updates the docs and the showcase app to have a good practice example of multiple new tab links together with a single warnig above.

## PR Checklist


cjrace:
Used my custom review agent locally and it passes all review checks cleanly. Will merge if CI passes too.

PR #250, opened 2026-09-12
Add external link icon option

cjrace:
## Overview of changes

Follow-up to [#195](https://github.com/dfe-analytical-services/shinyGovstyle/issues/195), and #249, add_warning now accepts "icon" alongside TRUE/FALSE.

It shows a small decorative arrow icon instead of the "(opens in new tab)" text, for grouped links where repeating the text isn't wanted but a visual cue still is. Folded into add_warning rather than a separate icon argument so text-and-icon can't both be shown at once. TRUE/FALSE behaviour is unchanged.

image

Pulls the icon from w3.org.

## PR Checklist


cjrace:
Used my custom review agent locally and it passes all review checks cleanly as well as my general poking around. Will merge if CI passes too.

PR #251, opened 2026-09-12
Add a default option for the banner function

### LINKED ISSUE #178: Add default text into the `banner()` function

cjrace:
## Overview of changes

Resolves #178 by adding a default option where you just supply a URL or mailto: link to the phase banner and get standard text.

## PR Checklist


cjrace:
Noticed the mailto: link wasn't using external_link(), will look into whether it can be updated to take mailto: links more easily / obviously.

PR #252, opened 2026-09-12
Add properly controlled widescreen layout

### LINKED ISSUE #98: Widescreen option
Hi, love this package, thank you for creating it 😊
Would it be possible to add a widescreen option please, so that the page looks more like a dashboard?
I've currently hacked this by editing .govuk-width-container (max-width of to 90vw and margin-right & margin-left to 0px) and .govuk-header__logo (padding-left: 15px) in govuk-frontend-norem.css.
Would prefer if the banner went all the way across the screen but I can't work out how to do that with my limited CSS knowledge.

### LINKED ISSUE #188: [Feature]: Function for adding standard metadata
### Description

Not fully thought through yet, but a `shinyGovstyle::site_metadata()` function might be helpful to guide developers to include key elements like setting the lang attribute in their apps.

### Use case

As a developer, I want a standardised function that guides me in setting the key metadata for my app, so that I don't have to do the research myself (and duplicate that with every other R Shiny app developer in gov).

### Why this is important

Elements like the lang attribute are important for accessibility, and this will also help to save time for users of the package, guiding them into the standard things they need to set

### Additional Context

Worth doing some research on what should / could be included here.

For valid language codes see [W3docs’ ISO Language Codes](https://www.w3docs.com/learn-html/html-language-codes.html).

### LINKED ISSUE #201: [Bug]: Remove `full_width_overrides()`

cjrace:
Long awaited, much asked for, and probably quite a beast to review... Grab a drink then have a read through before getting stuck in!

## Why does this PR exist?

This PR adds page-width control to all of shinyGovstyle, closes #201, closes #98, and closes #188.

Until now every shinyGovstyle app was locked to GOV.UK's standard ~960px content width, or CSS hacked with the ugly `full_width_overrides()` (I can call it ugly as I wrote it!).

The GOV.UK defaults didn't really work for dashboards and there's good cases for why dashboard design is different to most other services. Several users asked for a way to opt into a wider layout without having to hand-roll custom CSS.

This PR adds a width option to the components that need it, plus a new `gov_page()` wrapper that lets you set it once for the whole app instead of repeating it everywhere. This new page wrapper also sets the site metadata resolving #188.
 
Along the way it also cleans up some inconsistencies in the docs and examples.

### New width control
- A width argument (takes one of "standard" / "three-quarters" / "full" / a custom CSS length e.g. "1400px") has been added to every component that renders its own page-width container: `header()`, `footer()`, `banner()`, `cookieBanner()`, `service_navigation()`, and the main layout functions. 

- A new gov_page() function wraps a whole app and lets you set a default width once, which every component inherits unless it explicitly overrides it.

### New header spacing method

The bane of our existence!

- Now uses a mirror of the org name box on the right side of the header to keep centred before wrapping, seems to work for ensuring a genuinely centred title while also forcing a wrap underneath before it overlaps.

### Bug fixes

- Fixed an overlap between the organisation name and service name in the header, which could happen on wide screens at standard width. They now wrap onto separate lines gracefully instead of colliding.

- Removed a leftover gutter (extra side padding) that was still showing up at standard width even though it shouldn't have been there.

### Showcase app updates
- The example app now has a live width switcher under the phase banner, so you can toggle between standard, three-quarters and full width and see the effect immediately. 

- The intro text on that page has also been tidied up and attempts to make it clear this toggle only exists in the showcase, it's a custom toggle using JS to hack the showcase app, and isn't something available in the package. I based it on the trial language toggle component - https://design-system.service.gov.uk/components/language-navigation/.

### Documentation

- The layout vignette has a new section explaining the difference between page width (`gov_page()`) and content width (the existing two-thirds / one-third column system), including guidance on when to use each, with the Cookies tab in the showcase app as a live example of narrow content on a wide page.

- NEWS.md and roxygen docs updated throughout to describe the new argument and the two bug fixes above.

### Function example updates
- A handful of @examples blocks that used bare shiny::tags$br() calls purely as vertical spacing have been swapped for gov_text("Placeholder text"), so the examples only demonstrate real shinyGovstyle components rather than raw Shiny HTML tags.

- `gov_page()` has been swapped in for all examples as a clean and consistent approach that hopefully makes them less intimidating for users.

## PR Checklist

## Reviewer instructions

This one's partly about feel, not just correctness, so please:

- Run the showcase app locally and try the width switcher at each setting (standard / three-quarters / full). 
- Resize the browser window while you're at it, and keep an eye out for anything that looks "wonky": overlapping text, odd gaps, content that doesn't wrap the way you'd expect, etc.
- Read through the updated layout vignette and check the new "content width vs page width" explanation actually makes sense on a first read. Imagine you know nothing about the package when you're reading it.
- Skim the updated @examples blocks (?header, ?footer, ?gov_page, ?backlink_Input, ?select_Input are good ones to check) and confirm they're still easy to follow and actually runnable.
- Check the new width argument docs on a few different components, since they all now come from the same shared source. Would you know what to do if it was the first time you'd seen it?

Probably more too, though basically, this touches a lot of stuff and a lot of stuff that isn't tested as it's all visual, so give it a good thorough poking!


sarahmwong (approved):
Poking complete! Looks really good thank you for fixing this one Cam - happy that the functionality works well, is explained thoroughly in the vignette and workarounds with gov_box are clear and work nicely.

PR #253, opened 2026-09-13
Add GOV UK brand pkgdown theme

cjrace:
## Overview of changes

Been having a play with a _brand.yml GOV.UK theme for package documentation and quick styling - https://github.com/cjrace/govuk-brand-yml.

This branch shows what it would be like if we pulled it into shinyGovstyle.

image

## PR Checklist

## Reviewer instructions

Test that the site builds locally using `devtools::build_site()` and give it a poke around to view the styling.

Look over the comments included to see if it makes sense / you can understand what I've done and why.


sarahmwong (changes requested):
very neat! Is it possible to change the secondary/success/danger etc colours to the gov.uk defaults in the yml just so its all in line?

cjrace:
> very neat! Is it possible to change the secondary/success/danger etc colours to the gov.uk defaults in the yml just so its all in line?

Good catch @sarahmwong - hadn't clocked I'd pulled the secondary text colour instead of the secondary button colour 🤦 

Success / danger were using GOV UK colours, but I've done more to update the styling of the buttons so they're a bit more GOV UK like. It's difficult for some as there's no equivalent, like warning, which I've pulled out the GOV UK orange for. 

Fixed a few other small colour bugs I noticed as I went through too 😀 I've made sure both this branch and the separate repo have stayed up to date so that repo is a reusable version of this.

Let me know what you think!

